PR #1, opened 2023-11-19
feat(scraper): Setup Python env for scrapers

nattiechan:
This PR is based on Jiecheng's branch `jiedong111/cdktest`.

Docker container is not conducive to running Playwright in headful mode (i.e. browser pops up locally for visual checks). While a solution potentially exists (e.g. [this](https://github.com/piercefreeman/docker/tree/main/headfull-chromium) but is written for Node.js), getting that set up will take a bit of time.

In the meantime, we will utilize Python's virtual environments to isolate scraper development without polluting the local environment too much.

This PR also includes the following:

- Instructions on how to setup the environment locally
- A sample scraper and instructions on how to test them locally


jiedong111 (approved):
Looks good to me! Was able to run scraper after setting up the environment without issues. My PR for initial CDK setup is up as well.

nattiechan:
I have downgraded to Python 3.11, and further improved the scrapers to have some common methods to help simplify some of the logic. I would appreciate a re-review.

If you have previously tested this PR using Python 3.12, consider:

- Remove all virtual environments: `pipenv --rm`
- Remove Python 3.12 from `pyenv` and reinstall 3.11. You may also need to uninstall and re-install pipenv and pip during the process.
- Set python to 3.11: `pipenv global 3.11`

PR #2, opened 2023-11-19
feat(infra): Initial setup of CDK repo, basic Lambda DDB SNS

jiedong111:
This PR includes the initial setup of our CDK repo along with some base IaC including some Lambda functions, a state machine, and a dynamodb table. This should give an overview of how to interact with CDK resources across various stacks that we create to separate concerns when it comes to different facets of our application.

In order to provision resources locally and test infrastructure, you'll have to install CDK and bootstrap it with your AWS environment. You can take a look at the instructions in README.md for a more detailed overview on how to do so.


lib/database-stack.ts
```diff
@@ -0,0 +1,22 @@
...
+import * as dynamodb from 'aws-cdk-lib/aws-dynamodb';
+
+export class DatabaseStack extends cdk.Stack {
+  public readonly tableName: string;
+
+  constructor(scope: cdk.App, id: string, props?: cdk.StackProps) {
+    super(scope, id, props);
+
+    const hashTable = new dynamodb.Table(this, 'hashTable', {
+        partitionKey: {name: 'bank_name', type: dynamodb.AttributeType.STRING},
```
nattiechan, line 14: Will this be the same partition key that we discussed in the meeting? I forgot what it is exactly but I think it is more than just the bank name?

lib/shared-resources.ts
```diff
@@ -0,0 +1,158 @@
...
+    //     `cron-scraper-infra-${this.stackName}-lambda-execution-role`, 
+    //     {
+    //         assumedBy: new iam.ServicePrincipal("lambda.amazonaws.com"),
+    //         roleName: `cron-scraper-infra-${this.stackName}-lambda-execution-role`,
+    //         description: "Lambda function execution role that allows function to write to CloudWatch",
+    //         inlinePolicies: [
+    //             policyDocument,
+    //         ]
+    //     }
+    // );
```
nattiechan, line 61: Do we need this code?

lib/shared-resources.ts
```diff
@@ -0,0 +1,158 @@
...
+    
+    const scraperHandler = new lambda.Function(
+        this,
+        `cron-scraper-infra-${this.stackName}-lambda-execution-role`,
+        {
+            runtime: lambda.Runtime.PYTHON_3_9,
```
nattiechan, line 67: I am using 3.12 in my PR. Would that be an issue? Or you want me to downgrade? (prob prefer 3.10 minimum?)

lib/shared-resources.ts
```diff
@@ -0,0 +1,158 @@
...
+                "BANK": 'Bank of Kenya',
+            }
+        }
+    );
+
+    const validationHandler = new lambda.Function(
+        this,
+        `cron-scraper-infra-${this.stackName}-validation-handler-role`,
+        {
+            runtime: lambda.Runtime.PYTHON_3_9,
```
nattiechan, line 81: Same here

lib/shared-resources.ts
```diff
@@ -0,0 +1,158 @@
...
+        this,
+        `cron-scraper-infra-${this.stackName}-lambda-execution-role`,
+        {
+            runtime: lambda.Runtime.PYTHON_3_9,
+            code: lambda.Code.fromAsset("src"),
+            handler: 'scraper.main',
+            timeout: cdk.Duration.seconds(300),
+            environment: {
+                "BANK": 'Bank of Kenya',
+            }
```
nattiechan, line 73: I have not read the docs much, and I am wondering how this `environment` variable is used? As we have more and more scrapers, how would it affect the map here?

lib/shared-resources.ts
```diff
@@ -0,0 +1,158 @@
...
+        {
+            runtime: lambda.Runtime.PYTHON_3_9,
+            code: lambda.Code.fromAsset("src"),
+            handler: 'validation.main',
+            timeout: cdk.Duration.seconds(300),
+            environment: {
+                //Table names are unique in account per region, use table name to refer to table for operations
+                "TABLE_NAME": table.tableName,
+                "BANK": 'Bank of Kenya',
+            }
```
nattiechan, line 89: Do we plan to use one table per bank? Or a table by usage? (i.e. one table for notification, one table for scraper runs... etc)

lib/shared-resources.ts
```diff
@@ -0,0 +1,158 @@
...
+            environment: {
+                "TOPIC_ARN": userTopic.topicArn,
+                "BANK": 'Bank of Kenya'
+            }
+        }
+    )
+
+    const scrapeCircular = new tasks.LambdaInvoke(this, "Scrape Circular", {
+        lambdaFunction: scraperHandler,
+        resultPath: '$.hashedUrl',
```
nattiechan, line 110: Do we plan to store the URL in the DB? Or just in the scraper file is suffice?

lib/shared-resources.ts
```diff
@@ -0,0 +1,158 @@
...
+    const definition = sfn.Chain
+        .start(scrapeCircular)
+        .next(validateDatabase)
+        .next(new sfn.Choice(this, 'Validated?')
+            .when(sfn.Condition.booleanEquals('$.validated.validated', true), notifyUsers)
+        )
+
+    const scraperStateMachine = new sfn.StateMachine(this, 'ScraperStateMachine', {
+        definition,
+        timeout: cdk.Duration.seconds(300),
```
nattiechan, line 144: What is the max timeout we can have here?

lib/shared-resources.ts
```diff
@@ -0,0 +1,158 @@
...
+        tracingEnabled: true,
+        stateMachineType: sfn.StateMachineType.EXPRESS
+    })
+
+    rule.addTarget(new targets.SfnStateMachine(scraperStateMachine));
+
+    // The code that defines your stack goes here
+    new s3.Bucket(this, 'testbucket12341%13x', {
```
nattiechan, line 152: Should we put `'testbucket12341%13x'` or something along those lines in `.env` as well?

test/scraper-infra.test.ts
```diff
@@ -0,0 +1,17 @@
...
+//   const app = new cdk.App();
+//     // WHEN
+//   const stack = new ScraperInfra.ScraperInfraStack(app, 'MyTestStack');
+//     // THEN
+//   const template = Template.fromStack(stack);
+
+//   template.hasResourceProperties('AWS::SQS::Queue', {
+//     VisibilityTimeout: 300
+//   });
+});
```
nattiechan, line 17: If this test is not ready, let's remove the dead code for now

README.md
nattiechan, line 1: We may want to move the CDK-specific instruction into a different `.md` file, so we can use the `README` for more general epic-related matters

nattiechan (changes requested):
This looks great so far! Good work for setting this up! I can imagine this is not easy at all.

One overall comment - We need _a lot more comments/docstring_ on what each file and/or function is doing. You will be the subject expert in AWS CDK for this project, but we do not want to create a single point of failure so if you are not available, others can navigate around this infrastructure and make changes as needed.

Some of the things that stood out to me:

- files like validation.py and notification.py.. why do we need them? What do they actually do?
- `shared-recources.ts` could use a lot more comment/docstring, explaining what the functions are doing

lib/database-stack.ts
jiedong111: Yea, this ddb table setup was made prior to the meeting. I'm looking to get the partition key set straight sometime later today, but wanted to get this PR up so we can merge in the scraper stuff. I'll leave a comment on it in the code so we don't lose track of this

lib/shared-resources.ts
jiedong111: I left it in as a note for myself for setting up more complex role-based access control with CDK, but I'll remove it and just note it down somewhere for myself

lib/shared-resources.ts
jiedong111: It looks like Lambda only supports up to 3.11, so we'd have to downgrade from 3.12 to 3.11, I only provisioned these handlers with 3.9 to match the examples I was seeing since I was exploring things. I don't think running it in 3.11 would change anything

lib/shared-resources.ts
jiedong111: The environment variable for lambdas let you freely supply environment variables to the state of a particular lambda function, so this is just providing 'Bank of Kenya' as the bank for this specific scraper. We could literally hard code the relevant bank into the environment of each scraper handler if we needed to depending on our implementation

lib/shared-resources.ts
jiedong111: In a state machine, the timeout can be a year I think. We're going to transition to using just Lambda and Dynamodb streams to trigger other Lambdas, so our max timeout without using a state machine would be 15 minutes. 

If we find we need a longer timeout on some operation for some reason, we can transition that part specifically to a step function but step functions apply cost based on state changes and are much less generous than the free tier of lambda so if we can keep it to lambda/dynamo that would be best

lib/shared-resources.ts
jiedong111: I think based on the discussion we had with Alex, we're going to be storing:

> UUID,
> date_updated,
> hash_of_file,
> bank_name,

in the table and running with a composite key of UUID#hash_of_file and a range key of AWSTimeStamp

lib/shared-resources.ts
jiedong111: I think we're going to be using 1 dynamo table to store the bank data, use dynamo streams to trigger lambdas off changes to that table, then users would be informed with SNS and SQS. We'd have another microservice that would handle user subscriptions and consume from the events that we push from this microservice through SNS into SQS in order to send notifications.

lib/shared-resources.ts
jiedong111: I think test bucket is just a string that you provide that appends to the actual bucket name when CDK provisions it. Having that string exposed here isn't exposing any sensitive information

jiedong111:
Currently working through your comments and adjusting the PR, thanks for taking the time Nattie! 

'Validation.py' and 'notification.py' are additional lambda functions that were made as a proof of concept of setting up a step function. Since we're moving away from step functions they aren't strictly needed right now. I think they might be useful to keep around though just to emphasize that lambda functions should serve an individual purpose, so we have one function for scraping, one function for validating the scraped information against our table, and one function that sends a notification to SNS.

lib/database-stack.ts
nattiechan: Yep this is something that we can address in a follow-up PR

lib/shared-resources.ts
nattiechan: I will downgrade the environment to 3.11. Let's set that for AWS

lib/shared-resources.ts
nattiechan: Got it. Let's move `{"BANK": 'Bank of Kenya'}` example to a comment above L71 here and change the environment back to what we expect for now (probably `{}`?)

lib/shared-resources.ts
nattiechan: This may warrant further investigation re: price etc, but we may want to look into the possibly of having a table that has all executed scraper runs (both success or fail), and then another table that contains the information of the last success run and the `md5hash` of the file. I wonder if the cost of the lookup would be less and the cost of write is negligible

lib/shared-resources.ts
nattiechan: Let's set this to 5 minutes for now then. Though a comment on 300 seconds = 5 minutes and the max is 15 mins would be helpful

lib/shared-resources.ts
nattiechan: Okay. Eventually we may want to have a more "official" name other than test bucket in there

CDK.md
```diff
@@ -0,0 +1,71 @@
...
+* `npm run test`    perform the jest unit tests
+* `cdk deploy`      deploy this stack to your default AWS account/region
+* `cdk diff`        compare deployed stack with current state
+* `cdk synth`       emits the synthesized CloudFormation template
+
+## How to make it work
+
+Install CDK globally and bootstrap your environment, you'll need to provide CDK with necessary credentials to allow it to provision resources for your environment. In my case I am currently simply using my own training account's credentials, but eventually project specific credentials may be exposed for the people that are working on this infrastructure. Till then use your own settings.
+
+I'd recommend following the
```
nattiechan, line 24: typo?

CDK.md
```diff
@@ -0,0 +1,71 @@
...
+
+Currently the stacks with actual definitions are database stack which contains a simple DDB table and shared-resources stack which has a variety of resources related to Lambda and SNS.
+
+The shared-resources stack references logical resources contained in the database-stack, so provision the database-stack first using 'cdk deploy DatabaseStack'r changes and compile
+* `npm run test`    perform the jest unit tests
+* `cdk deploy`      deploy this stack to your default AWS account/region
+* `cdk diff`        compare deployed stack with current state
+* `cdk synth`       emits the synthesized CloudFormation template
+
+## How to make it work
```
nattiechan, line 20: For a follow-up PR: consider added a step-by-step instruction, or a link to said information, on how to set things up

bin/scraper-infra.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const dbStack = new DatabaseStack(app, 'DatabaseStack');
+
+const sharedResources = new SharedResourcesStack(app, 'ScraperInfraStack', {
+  /* If you don't specify 'env', this stack will be environment-agnostic.
+   * Account/Region-dependent features and context lookups will not work,
+   * but a single synthesized template can be deployed anywhere. */
+
+  /* Uncomment the next line to specialize this stack for the AWS Account
+   * and Region that are implied by the current CLI configuration. */
```
nattiechan, line 17: We uncommented the next line already

```suggestion
  /* Specialize this stack for the AWS Account
   * and Region that are implied by the current CLI configuration. */
```

lib/database-stack.ts
```diff
@@ -0,0 +1,23 @@
+import * as cdk from 'aws-cdk-lib';
+import * as dynamodb from 'aws-cdk-lib/aws-dynamodb';
+
+//stack for database, RDS, DDB tables etc
```
nattiechan, line 4: Dumb question: what is a stack?

nattiechan:
> 'Validation.py' and 'notification.py' are additional lambda functions that were made as a proof of concept of setting up a step function. Since we're moving away from step functions they aren't strictly needed right now. I think they might be useful to keep around though just to emphasize that lambda functions should serve an individual purpose, so we have one function for scraping, one function for validating the scraped information against our table, and one function that sends a notification to SNS.

I am not a huge fan of leaving dead code in the main branch. One thing you can do to preserve the information is to have a separate branch in this repo, put the code where they should go with some information in the commit on how the information should be consumed.

Of course, another way is to have a separate markdown that logs your findings on how to set up step function (that would be amazing too!)

lib/database-stack.ts
jiedong111: Since CDK compiles into Cloudformation templates, it uses a lot of similar syntax. A stack in CDK corresponds to a stack in CloudFormation which is basically a bundle of logical resources that CloudFormation takes and translates into/provisions actual physical infrastructure for.

jiedong111:
My concern with removing 'scraper.py', 'validation.py', and 'notification.py' is that without them the state machine currently represented in the code won't actually do anything. They're the lambda functions that are referenced by the current lambda handlers defined in 'shared-resources-stack'. I probably should have been more clear in that they're not strictly necessary for what we'd be doing in the future, but they are necessary for the code as is to run properly once you deploy

lib/shared-resources.ts
nattiechan: bump on this

lib/shared-resources.ts
nattiechan: this as well

src/notification.py
```diff
@@ -0,0 +1,8 @@
+import json
+import boto3
+import os
+sns_client = boto3.client('sns')
+
+def main(event, context):
+    response = sns_client.publish(TopicArn=os.environ.get('TOPIC_ARN'),Message="Test message")
+    return('success?')
```
nattiechan, line 8: Miss new line at the end of the file (Python linting standards). Same for all `.py` files in this PR

nattiechan (commented):
Would again, appreciate some comments/docstring in `notification.py` and `validation.py` on the purposes of the files

jiedong111:
> Would again, appreciate some comments/docstring in `notification.py` and `validation.py` on the purposes of the files

Python files should be adjusted, also updated the runtime on the lambda handlers. Tell me if anything else stands out!

lib/shared-resources.ts
```diff
@@ -0,0 +1,176 @@
...
+            }
+        }
+    )
+
+    //as above, but manages notifications
+    const notificationHandler = new lambda.Function(
+        this,
+        `cron-scraper-infra-${this.stackName}-notification-handler-role`,
+        {
+            runtime: lambda.Runtime.PYTHON_3_9,
```
nattiechan, line 102: Sorry I missed this - we will want to change this to 3.11 as well if possible

nattiechan (approved):
Not sure if you want Sabrina to take a look at this as well, and just one last thing before merging

jiedong111:
I think we should wait till @sgoldfarb2 signs off, adjusted the runtime for 'notification.py'.

CDK.md
sgoldfarb2: Definitely +1 to this, it definitely looks intimidating for someone just coming in

lib/database-stack.ts
sgoldfarb2: would you mind even adding a ticket for this too on the scraper project board @jiedong111 ? this way it can't get lost in all the new added work

sgoldfarb2 (approved):
this all looks great to me! my only additional (probably silly) question is that this is already setting up a real dynamoDB table, yes? So is there a situation where we could accidentally get ourselves in a pickle with $ if we had say...a hackathon to make 240 scrapers over a weekend and folks were running them often to test them?

jiedong111:
>this all looks great to me! my only additional (probably silly) question is that this is already setting up a real dynamoDB table, yes? So is there a situation where we could accidentally get ourselves in a pickle with $ if we had say...a hackathon to make 240 scrapers over a weekend and folks were running them often to test them?

The CDK repo will deploy actual resources, but it's configured and bootstrapped for someone's individual credentials. At some point if we have a lot of people working on the infrastructure, we'd have to set up a developer account or some kind of access control to give them access to an existing infrastructure. Since we're only setting things up right now, the way the repo is built it's intended that if an individual contributor wants to mess around with/test the infrastructure they would have to bootstrap their own environment and provide their own credentials. Our scraper local testing doesn't require any deployment to work though, so unless someone manages to bootstrap their environment and deploy accidentally they wouldn't incur any costs, and if they did it would be to their personal AWS accounts. I think it's worth making a bold warning on in CDK.md about that though, so people make sure to properly tear down any environments that they spin up if they go to the trouble of setting things up.

>would you mind even adding a ticket for this too on the scraper project board @jiedong111 ? this way it can't get lost in all the new added work

Just to clarify here, do you mean a ticket for setting up our dynamo tables? But I'd definitely be willing to set up a ticket for it in the scraper project board

nattiechan:
> So is there a situation where we could accidentally get ourselves in a pickle with $ if we had say...a hackathon to make 240 scrapers over a weekend and folks were running them often to test them?

@sgoldfarb2 as long as:

- The devs test scrapers locally per instructions in my PR
- The scrapers do not connect to AWS in any way (no read/write to/from DynamoDB and s3 bucket)

Then the only concern is devs accidentally bombard the website too much, which I already cautioned in the docs

PR #5, opened 2023-11-29
Feature/gh 68 bank of rwanda scraper

SFJohnson24:
created bank of rwanda scraper with some tweaks to original example.


scrapers/bank_of_rwanda.py
```diff
@@ -0,0 +1,46 @@
...
+BASE_URL = "https://www.bnr.rw"
+CIRCULAR_URL = f"{BASE_URL}/laws-and-regulations/banking/directives-guidelines/"
+
+
+def bank_of_rwanda_scraper():
+    with sync_playwright() as pw:
+        browser = get_browser(pw)
+        page = browser.new_page()
+        page.goto(CIRCULAR_URL)
+        print('circ url ', CIRCULAR_URL)
```
nattiechan, line 14: Please remove these print statements that are more for debugging purposes

scrapers/bank_of_rwanda.py
```diff
@@ -0,0 +1,46 @@
...
+        print('pdf_link ', pdf_link)
+        link = pdf_link.get_attribute("href")
+        print('link ', link)
+        absolute_url = f"{BASE_URL}/{link}"
+        print('absolute_url ', absolute_url)
+        # If you need to get the file via a request, then you will need to
+        # save the contents from the request to a file as shown below
+        # If you need to interact with a certain button to download the file
+        # then consider incorporate Playwright's download capability:
+        # https://playwright.dev/python/docs/downloads
```
nattiechan, line 29: If your scraper works with how things are in L32, then we can remove this comment

scrapers/bank_of_rwanda.py
```diff
@@ -0,0 +1,46 @@
...
+def bank_of_rwanda_scraper():
+    with sync_playwright() as pw:
+        browser = get_browser(pw)
+        page = browser.new_page()
+        page.goto(CIRCULAR_URL)
+        print('circ url ', CIRCULAR_URL)
+
+        session = requests.Session()
+        # Only the most recent circular is considered
+        # TODO: Option to download n number of files
```
nattiechan, line 18: We will keep the scope to the scrapers to download the most recent one only. Feel free to remove this `TODO` and also the same `TODO` in the example scraper (Keep the comment in L17 though please)

nattiechan (changes requested):
Please make sure your PR is pointed to `main` as the base, not `dev`. Feel free to look up on google on how to do so.

scrapers/bank_of_rwanda.py
SFJohnson24: I redirected my PR towards main and have removed the comments, print statements.  Also merged recent changes from main into the feature branch.

scrapers/bank_of_rwanda.py
```diff
@@ -0,0 +1,35 @@
...
+
+def bank_of_rwanda_scraper():
+    with sync_playwright() as pw:
+        browser = get_browser(pw)
+        page = browser.new_page()
+        page.goto(CIRCULAR_URL)
+
+        session = requests.Session()
+        pdf_link = page.locator("td > a").all()[0]
+        link = pdf_link.get_attribute("href")
```
nattiechan, line 17: From a brief look at the website's DOM this is a touch fragile, because `td > a` is not the most specific thing in the world. I can have the following on the same page, but one link points to the files that we want and another not - 

```
<td class="foo">
  <a href="some-link">
....
```

```
<td class="bar">
  <a href="some-link-2">
....
```

and I cannot guarantee `page.locator("td > a").all()[0]` would give me exactly what we are looking for.

We cannot completely eliminate the risk, though we can mitigate it a few different ways. In this context I would try to include a few more details so the xpath becomes more specific.

I have not fully tested this, though try to change the xpath to `table.table > * > tr > td > a` and see if you can get a similar result.

nattiechan: Also, from a defensive programming stand point, part of me starts to wonder if we should check and make sure `link` contains some value before continuing.

nattiechan (changes requested):
looks a lot better already!

scrapers/bank_of_rwanda.py
qpgdev: Nattie, can you go into a little more detail about what you mean by "make sure link contains some value"? Are you asking us to confirm that link !== null && link.length !== 0? In addition, can you help me understand the significance of doing this?

nattiechan: So let's go through this together.

> can you go into a little more detail about what you mean by "make sure link contains some value"? Are you asking us to confirm that link !== null && link.length !== 0?

First, do you know what we expect `link` to be? If you are unsure, try running the scraper in debug mode using VS code so you can see the value of `link`.

And if the `link` does not contain some value, how would it affect the rest of the code?

qpgdev: Here's what we've come up with:

```
        # Check if the link is type string and if last four chars are .pdf (index slicing)
        if type(link) == str and link[-4:] == '.pdf':
            response = session.get(link)
            with download_filepath(__file__) as filepath:
                    file.write(response.content)

            if parse_arguments().debug:
```
If this looks sufficient, we'll go ahead and modify it!

nattiechan: Cool! Definitely a step in the right direction.

Consider this:

With this change, the scraper will only download if `link` passes the checks. If it does not, then the scraper does nothing. In both scenarios, the script will finish running with no errors. How could we determine if a successful scraper run has the correct link and, therefore, a file download and one that does not without fancy and/or complicated monitoring?

What if I told you that the mechanism that runs this determines if a scraper runs successfully or not based on whether there is an error raised during the process?

How would that information change your approach?

Hint: look up `how to raise an exception python`

SFJohnson24: Hi Nattie,
We added a try/except block and are throwing exceptions for both type and value errors that might occur.  Let Quinn and I know what you think.  We were not certain if the try/except was necessary if there was already external error handling in the program running each scraper file.
Sam

nattiechan: The exception looks good.

Though I am still looking for some resolution on my original comment (a separate issue): https://github.com/sgoldfarb2/scraper-infra/pull/5/files#r1410203304

scrapers/utils.py
```diff
@@ -97,3 +98,8 @@ def get_md5hash(filepath):
             md5.update(data)
 
     return md5.hexdigest()
+
+def is_pdf(url):
+    response = requests.head(url)
+    content_type = response.headers.get('Content-Type', '')
+    return 'application/pdf' in content_type
```
nattiechan, line 105: 2 things:

1. Instead of breaking up into 2 checks, I would probably combine the check for null value here, with the _option_ to also check if it is a PDF as well (because I don't think we should assume it would be PDF for all circulars)

2. Please make sure there is an empty line at the bottom of a python file

```suggestion
def validate_and_get_download_link(base_url, download_link, check_content_pdf=True):
    # Add your check for null here, and raise if it is invalid
    is_valid_link = .....
    absolute_url = f"{base_url}/{download_link}"
    if check_content_pdf:
      response = requests.head(absolute_url)
      content_type = response.headers.get('Content-Type', '')
      is_pdf = 'application/pdf' in content_type
      if not is_pdf:
          raise ....
    return absolute_url
```
Then in scrapers, you can do:

```
validate_and_get_download_link(BASE_URL, link)
```

scrapers/bank_of_rwanda.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def bank_of_rwanda_scraper():
+    with sync_playwright() as pw:
+        browser = get_browser(pw)
+        page = browser.new_page()
+        page.goto(CIRCULAR_URL)
+
+        session = requests.Session()
+        pdf_link = page.locator("div.row > table > tbody > tr > td > a >> nth=0")
```
nattiechan, line 16: I know I am being a bit annoying here. Again similar to what I mentioned in a different PR we don't want to have the XPath go down so many layers if possible. Moreover, all these tags do not have anything specific to them (except for the top) so we run into the same issue of this is not as specific as you think it is.

Curious if `table.table > * > tr > td > a` worked?

Also, since we will transition this to get latest X circulars later let's not specify `nth=0` so we can make the transition easier

nattiechan (changes requested):

scrapers/bank_of_rwanda.py
SFJohnson24: ```pdf_link = page.locator("table.table > * > tr > td > a") ``` did not work but when I tweaked it to ```pdf_link = page.locator("table.table > * > tr > td > a").all()[0]``` it did get the PDF

nattiechan (approved):
Great work!

gizmo572 (approved):
Looks good. I tested it and saw no issues.

PR #6, opened 2023-11-29
docs(contrib): update contributing docs

jpettis1:
Based on feedback for environment setup and common issues, updating the environment setup steps to include resolving playwright dependency issues and Python interpreter setup in VS Code.


CONTRIBUTING.md
```diff
@@ -43,6 +43,12 @@ In the terminal:
 - Start a virtual environment: `$ pipenv shell`
   - If `pipenv` requires you to specify a specific Python version: `$ pipenv shell --python 3.11`
 - Install dependencies: `$ pipenv install`
+- Check to make sure playwright is installed in your virtual envionrment: `$ pipenv run pip list`
```
nattiechan, line 46: typo

```suggestion
- Check to make sure playwright is installed in your virtual envionment: `$ pipenv run pip list`
```

also - can you specify what folks should be looking for - `pytest-playwright`? or `playwright` on the list? (may want to clarify that)

CONTRIBUTING.md
```diff
@@ -43,6 +43,12 @@ In the terminal:
...
+  - If playwright is not installed run: `$ playwright install`
+- Select the correct Python interpreter in VS Code to run your scraper code:
+  - Navigate to the command palette in VS Code
+  - Select `Python: Select Interpreter`
+  - Select the interpreter referenced in the project's virtual environment:`Python 3.11.6('scraper-infra...') <path> PipEnv`
```
nattiechan, line 51: Can you move this to below the paragraph, after L53? I don't think we should assume everyone is going to be using VS code

nattiechan (changes requested):
Thank you! Just a couple things

CONTRIBUTING.md
jpettis1: Absolutely. This is complete and pushed.

CONTRIBUTING.md
jpettis1: Yes, that makes complete sense. This is now pushed.

nattiechan (approved):
Thanks!

willhack (approved):
LGTM 🚀

PR #7, opened 2023-11-29
Feature/gh 29 bank of zambia scraper

### LINKED ISSUE #29: [SPIKE] Error handling in Lambda functions/serverless #1
Leaving initial ticket here to note this as an active issue. 

Looking to investigate how exactly errors bubble up through Lambda functions. We're looking into potentially introducing step function workflows as well, which have their own redrive functionality. How do errors bubble up within state machines, and what level of control do we have over error handling there?

What are best practices for error handling in serverless?

We have a several queues in our architecture as well, how does error handling change when a Lambda function is expected to push to or is triggered by a queue? How are we going to manage our DLQs, what are sensible retries/throughput?

staclb:
Created bank of zambia scraper based on existing kenya example. 
Resolves #29


nattiechan:
Nice work! The fun challenge becomes - are you confident walking through what the scraper is _actually_ doing? ;)

staclb:
> Nice work! The fun challenge becomes - are you confident walking through what the scraper is _actually_ doing? ;) Why is the scraper build the way it is? What is with all the `with .....`?

scrapers/bank_of_zambia.py
```diff
@@ -0,0 +1,33 @@
...
+def bank_of_zambia_scraper():
+    with sync_playwright() as pw:
+        browser = get_browser(pw)
+        page = browser.new_page()
+        page.goto(CIRCULAR_URL)
+        session = requests.Session()
+
+        pdf_link = page.locator("div.list-group > a").all()[0]
+        link = pdf_link.get_attribute("href")
+        absolute_url = f"{BASE_URL}{link}"
```
nattiechan, line 18: While reviewing https://github.com/sgoldfarb2/scraper-infra/pull/5, I realized that perhaps it may be good to double-check to ensure `link` has value before continuing. Could we add a check here and also a check in `bank_of_kenya` scraper?

nattiechan (changes requested):

scrapers/bank_of_zambia.py
staclb: @nattiechan Implemented the error handling for links with exception handling in mind

scrapers/bank_of_kenya.py
```diff
@@ -17,6 +17,12 @@ def bank_of_kenya_scraper():
         # TODO: Option to download n number of files
         pdf_link = page.locator("td.pdf_link > a").all()[0]
         link = pdf_link.get_attribute("href")
+
+        # Check for link validity
+        if link is None or not link.endswith(".pdf"):
+            browser.close()
+            raise ValueError("Invalid link or not found")
```
nattiechan, line 24: Consider the following.

```
link = None
bool(link is None) ---> # True

link = ''
bool(link is None) ---> # False
```

In this case we want to ensure that if we don't have a link we need to stop the script. Moreover, in this case specifically, an empty string should not pass the check either. So we should do:

```suggestion
        if not link or not link.endswith(".pdf"):
            browser.close()
            raise ValueError("Invalid link or not found")
```

or if we want to go a step further:

```suggestion
        if not link or (isinstance(link, str) and not link.endswith(".pdf")):
            browser.close()
            raise ValueError("Invalid link or not found")
```

They both have pros and cons tbh. I will leave you to figure out what the second one may miss that the first one won't, and vice versa.

Though I think for all intents and purposes, let's go with the first suggestion and be more restrictive if we want.

nattiechan (changes requested):

scrapers/bank_of_kenya.py
staclb: Thank you for the insights, I went ahead and implemented the first case, and I agree that a more strict approach can be introduced later if need be.

scrapers/bank_of_zambia.py
```diff
@@ -0,0 +1,39 @@
...
+def bank_of_zambia_scraper():
+    with sync_playwright() as pw:
+        browser = get_browser(pw)
+        page = browser.new_page()
+        page.goto(CIRCULAR_URL)
+        session = requests.Session()
+
+        pdf_link = page.locator("div.list-group > a").all()[0]
+        link = pdf_link.get_attribute("href")
+        link = ""
```
nattiechan, line 18: I am guessing this is from you testing? By doing this you are overriding the `link` variable and this scraper should be consistently hitting the error below which is not what we want

scrapers/bank_of_zambia.py
```diff
@@ -0,0 +1,39 @@
...
+
+        if not link or not link.endswith(".pdf"):
+            browser.close()
+            raise ValueError("Invalid link or not found")
```
nattiechan, line 22: Since you will need to make some changes anyway, let's clean up the error message here and in the other scraper as well to make it clearer.

```suggestion
        if not link or not link.endswith(".pdf"):
            browser.close()
            raise ValueError("Invalid file URL.")
```

nattiechan (changes requested):
You may want to re-test both scrapers to ensure they work as expected

nattiechan (approved):
Nice work!

jiedong111 (approved):
Great work, looks good to merge!

PR #8, opened 2023-12-01
Jiedong111/cdkextension

jiedong111:
# Workflow Implementation for Event-Driven Architecture

This PR introduces a workflow that integrates various AWS services using an event-driven architecture. The workflow is as follows:

## Workflow Steps

1. **State Machine Subscription to EventBridge Cron Event**
   - The state machine is triggered by a cron event initiated by EventBridge.

2. **Initial Population Function**
   - The state machine executes an initial population function to populate values into `uuid` and `scraper-run` tables.

3. **Mock Scraper Function Execution**
   - Executes a mock scraper function, which sends a message with specific attributes to the scraper SQS queue.

4. **Validation Process**
   - `validation.py` consumes the message from the queue, compares the hash-content of the message's body, adds additional necessary attributes, and then puts the item into the `scraper-run` table.

5. **Stream Reader Activation**
   - `stream-reader.py` is subscribed to the DynamoDB stream originating from the `scraper-run` table. It reads every operation in the table and looks for "INSERT" events. Upon finding an "INSERT" event, it checks if the 'notified' attribute is set to 'True'. If so, it sends a message to the notification SQS queue.

6. **Notification Sending**
   - `notification.py` sends a notification to the email address subscribed to the `userTopic` in `shared-resource.ts`.

## Additional Information

- **Comments and Documentation**: Each of the Lambda handlers and the infrastructure files have additional comments

- **Current Testing Focus**: Currently, the workflow is set to test only the 'notify' path of the microservice. The `scraper.py` forces a new hash every time the workflow runs by appending its `time_stamp` to `hash_content`.

- **Error Handling**: Further investigation into error handling with Lambda is planned, but this PR hopefully provides a stable merge point for the current state of the workflow.

- **Potential Integration of Actual Scraper**: This setup paves the way for potentially running an actual scraper on AWS rather than using a mock scraper.

- **Note on `cdk watch` Debugging**: Previously, while debugging `cdk watch`, all python/local scraper testing components were removed from my branch because I thought they might be messing with `cdk watch`. I'll add them back in manually if need be but Github seems to think this is fine to merge so I'm not sure? Would love some input on this if people have been in a similar situation

## Next Steps

- Review and merge this PR as a stable point for the workflow.
- Plan the integration of actual scraper services in AWS.
- Enhance error handling capabilities within the Lambda functions.


jiedong111:
Closing this PR to allow for PR with re-factored infra

PR #10, opened 2023-12-04
Feature/gh 70 bank of ghana scraper

SFJohnson24:
created bank of ghana scraper


scrapers/bank_of_ghana.py
```diff
@@ -0,0 +1,37 @@
...
+
+def bank_of_ghana_scraper():
+    with sync_playwright() as pw:
+        browser = get_browser(pw)
+        page = browser.new_page()
+        page.goto(CIRCULAR_URL)
+
+        session = requests.Session()
+        
+        pdf_link = page.locator('div > div > div > div > div > section > div > div > div > div > div > section > div > div > div> div > div > div > div > h2 > a').all()[0]
```
nattiechan, line 16: Unfortunately, this xpath is too long and needs more precise. When there is no additional information besides the tag (i.e. `div`..etc.), We cannot guarantee we can get the exact hierarchy we are going after.

So this is the DOM that we are interested in:
image

Instead, I suggest identifying the most identifiable element and then working upward.

- Find the `Download` element via searching by role and text (or similar)
- Get the parent `a` tag, 2 levels up

Alternatively, you can try:
```
page.locator("//a[contains(@class, 'jet-button')]")
```
(Note: untested - you will need to verify if that works)

The first method may be slightly better because we can find a button by role and text, but if not good 'ol XPath should work as well

BONUS: If you can find all the `a` tag using this new approach and just get the first element from the list, which will help us down the road

nattiechan (changes requested):

scrapers/bank_of_ghana.py
SFJohnson24: Hi Nattie,
Quinn and I have refactored the path to use the jet-button class.  Additionally, we had to tweak the utils function we invoked to validate our link.  We believe it will now properly validates both PDFs and links to PDFs as best as possible.

scrapers/utils.py
```diff
@@ -111,7 +112,7 @@ def validate_and_get_download_link(base_url, download_link, check_content_pdf=Tr
         response = requests.head(absolute_url)
         content_type = response.headers.get('Content-Type', '')
         is_pdf = 'application/pdf' in content_type
-        if not is_pdf:
+        if not is_pdf and download_link[-4:] != '.pdf':
```
nattiechan, line 115: This is probably better

```suggestion
        if not is_pdf and download_link.endswith('.pdf'):
```

scrapers/utils.py
```diff
@@ -111,7 +112,7 @@ def validate_and_get_download_link(base_url, download_link, check_content_pdf=Tr
...
             raise TypeError('Content-Type is not application/pdf.')
```
nattiechan, line 116: This is probably better as well

```suggestion
            raise TypeError('File extension does not match content type.')
```

nattiechan (changes requested):
Also - can we possibly add some tests for `validate_and_get_download_link()`? You may need to look online on best practice on writing tests - especially around whether we should have a dedicated test folder to house the tests.

Side note: There is another [PR](https://github.com/sgoldfarb2/scraper-infra/pull/15) that has tests implemented. I was going to tell them to look up whether we need a test folder or not, but I think yours is further along so let's have you look up and set up the infrastructure first.

scrapers/utils.py
```diff
@@ -2,14 +2,15 @@
...
 import inspect
-import requests
 import shutil
 import sys
 import tempfile
 from contextlib import contextmanager
 from datetime import datetime
 from pathlib import Path
 
+import requests
```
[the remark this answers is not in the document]
qpgdev: Hey Nattie, we're running into issues where the requests import is being autoformatted to this position.  Unsure if there have been linting changes to main that I'm not aware of - we're working on resolving this ASAP.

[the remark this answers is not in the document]
nattiechan: Hmmm okay if `isort` does that we can just leave it as is for now. I would prioritize enabling the extension rather than trying to wrangle this import

nattiechan:
> Also - can we possibly add some tests for `validate_and_get_download_link()`? You may need to look online on best practice on writing tests - especially around whether we should have a dedicated test folder to house the tests.
> 
> Side note: There is another [PR](https://github.com/sgoldfarb2/scraper-infra/pull/15) that has tests implemented. I was going to tell them to look up whether we need a test folder or not, but I think yours is further along so let's have you look up and set up the infrastructure first.

UPDATE: @SFJohnson24, we still want to add tests, but let's wait till https://github.com/sgoldfarb2/scraper-infra/pull/15 merges in and then open a 2nd PR just for the tests.

So path for merging is:
- Let's clean up the above comments (don't worry about tests right now)
- get this reviewed, approved and then merge in

scrapers/utils.py
SFJohnson24: I just committed our changed--we did add testing and were about to make one last test mocking the headers to test a PDF and a header that is plain text for our function.  I stopped when I saw your comment on this PR but committed what we had to this PR.  We can remove our testing file and just merge what we have and transfer the tests to a new branch?  Let us know how you want us to proceed

nattiechan: Let's remove the testing file and do it in a separate PR, after https://github.com/sgoldfarb2/scraper-infra/pull/15 is merged

nattiechan:
@SFJohnson24 you will want to merge the latest `main` into your branch and resolve the conflicts

scrapers/utils.py
```diff
@@ -128,9 +128,9 @@ def validate_and_get_download_link(base_url, download_link, check_content_pdf=Tr
...
-        content_type = response.headers.get("Content-Type", "")
-        is_pdf = "application/pdf" in content_type
-        if not is_pdf:
-            raise TypeError("Content-Type is not application/pdf.")
-
+        content_type = response.headers.get('Content-Type', '')
+        is_pdf = 'application/pdf' in content_type
+        if not is_pdf and download_link.endswith('.pdf'):
+            raise TypeError('File extension does not match content type.')
+
```
nattiechan, line 135: Can we check if our linting enforces Python quotes as single or double quotes? I am looking at other scrapers and it looks like we are consistently using double quotes

SFJohnson24: there does not appear to be a linter for the scraper files as far as I can tell.  We could implement one in a separate PR if that is what you would like us to do?  We fixed the formation regardless.

scrapers/bank_of_ghana.py
```diff
@@ -0,0 +1,52 @@
...
+    with sync_playwright() as pw:
+        browser = get_browser(pw)
+        page = browser.new_page()
+        page.goto(CIRCULAR_URL)
+
+        session = requests.Session()
+
+        pdf_link = page.locator("//a[contains(@class, 'jet-button')]").all()[0]
+        link = pdf_link.get_attribute("href")
+        link = link.replace(BASE_URL, "")
```
nattiechan, line 27: Instead of overriding the variable

```suggestion
        link = pdf_link.get_attribute("href").replace(BASE_URL, "")
```

scrapers/bank_of_ghana.py
```diff
@@ -0,0 +1,52 @@
...
+
+        parsed_link = link.split("Hash=")[-1]
```
nattiechan, line 29: If this is only used once in L39 below, then we can just do

```
filename = get_standardized_filename(__file__, link.split("Hash=")[-1])
```

nattiechan (changes requested):
Just last couple of minor things

scrapers/utils.py
nattiechan: It will be based on the black formatted and the VS code settings here: https://github.com/sgoldfarb2/scraper-infra/blob/main/.vscode/settings.json

Do you have the extensions installed? If you are not using VS code then yeah we may need explicit linting rules for Python

SFJohnson24: I have black formatter installed on my VS code.  We were not getting any problems when using single quotes.  It may be black formatter interacting with another extension Prettier that also can have rules for quotations.

nattiechan (approved):
Make sure you re-test this before you merge :)

jpettis1:
Looks great and is working as expected! Good use of utility for ensuring we have a valid download link. Looking forward to seeing some tests in the next PR. Signing off on this!

jpettis1 (approved):

PR #12, opened 2023-12-04
add scraper for Bank of Malawi

gizmo572:
Utilize Playwright to interact with the webpage to download and save the most recent circular pdf file.
Resolves #31


scrapers/bank_of_malawi.py
```diff
@@ -0,0 +1,34 @@
...
+        with page.expect_download() as download_info:
+            page.click(f"tbody > tr:nth-child({1}) > td > a")
+
+        download = download_info.value
+
+        with download_filepath(__file__) as filepath:
+            download.save_as(filepath)
+            # TODO: verify file content before deciding whether we should send the file to s3 buckets
+            #       and the md5hash to DynamoDB
+            md5hash = get_md5hash(filepath)
```
nattiechan, line 25: I like that you are utilizing the playwright functionality for this. That said, once contributors build the 1st iteration of the scraper, we need to transition to having scrapers download the latest X number of circulars. Can we update this to allow an easy transition from 1 to 10 downloads?

You can accomplish that by getting the list of `a` tags, and then we will download either via the playwright functionality or request.

nattiechan (changes requested):

nattiechan:
Also - could you please link the issue that this PR addresses, similar to how other PRs are? Thanks!

scrapers/bank_of_malawi.py
gizmo572: hey Nattie, I had this working for multiple files by using the following syntax:
"""
        pdf_links_count = page.eval_on_selector_all("td > a", "elements => elements.length")
        count = min(n, pdf_links_count)

        for i in range(count):
            with page.expect_download() as download_info:
                page.click(f"tbody > tr:nth-child({i+1}) > td > a")

            download = download_info.value
            id = parse_link(download.url)

            with download_filepath(__file__, id) as filepath:
                download.save_as(filepath)
                # TODO: verify file content before deciding whether we should send the file to s3 buckets
                #       and the md5hash to DynamoDB
"""

My thought process is that the href in an 'a' tag here is not representative of the actual pdf file; it's just a link to trigger the download of the desired pdf file. We can get a more accurate reflection of the pdf file name by using the value of "download.url". If we do this, the only use case for retrieving the anchor tags is to get the number of anchor tags, so we just retrieve the count here to save space. What do you think about this? (also disregard the 'id' stuff)

nattiechan:
FYI @gizmo572 - looks like you will need to link this PR by selecting the PR from the `Development` section on the right hand side of the related issue. I just linked the PR to the issue for you

gizmo572:
The download object has a 'suggested_filename' property that is a string ending with the file's extension, so I'm using this to check if it's a .pdf file before saving. I tried using the 'requests.Session()' method, and essentially mimicked the current logic in the Rwanda scraper, but the 'Content-Type' property in the header is set to 'application/octet-stream', rather than 'application/pdf', so there is an error when validating the file extension in the 'validate_and_get_download_link' function. If we ignore this check, the 'requests.Session()' method works fine. An alternative method to validate the file extension when using this method is as follows:
```
            absolute_url = f"{BASE_URL}{link}"

            content_disposition = (
                response.headers.get("content-disposition")
                )
            parsed_CD = content_disposition.split('filename="')[-1].split('"')[0]

            if not parsed_CD.endswith('.pdf'):
              raise ValueError('Invalid file extension')
```

scrapers/bank_of_malawi.py
```diff
@@ -0,0 +1,54 @@
...
+        page = browser.new_page()
+        page.goto(CIRCULAR_URL)
+
+        pdf_links_count = page.eval_on_selector_all(
+            "table.table > * > tr > td > a", "elements => elements.length"
+        )
+        count = min(n, pdf_links_count)
+
+        for i in range(count):
+            with page.expect_download(timeout=10000) as download_info:
```
nattiechan, line 27: Is this in seconds? How long is this in minutes?

nattiechan:
> so there is an error when validating the file extension in the 'validate_and_get_download_link' function

That's fine - not every scraper can utilize that function. If this scraper cannot then just pick an appropriate approach for a similar check and go from there

scrapers/bank_of_malawi.py
nattiechan: Bump on this comment above regarding `page.click(f"tbody > tr:nth-child({i+1}) > td > a")`. Is it possible to utilize the `pdf_links_count` to get the URLs and get the file instead of having to repeat the XPath twice?

scrapers/bank_of_malawi.py
```diff
@@ -0,0 +1,54 @@
...
+                page.click(f"table.table > * > tr:nth-child({i+1}) > td > a")
+
+            download = download_info.value
+            suggested_filename = download.suggested_filename
```
nattiechan, line 31: Also bump on the comment here: https://github.com/sgoldfarb2/scraper-infra/pull/12#discussion_r1418390635

Would there be a situation where `download` is None` and then `download.suggested_filename` would error out?

scrapers/bank_of_malawi.py
gizmo572: This is milliseconds. 10 seconds. The default is 30 seconds. This is how long the program will wait for a download to happen before timing out.

nattiechan: I see. Is there a reason why we want to shorten the timeout by 20 seconds?

gizmo572: This is the amount of time to wait for the download to start. 30 seconds seems excessive given all we're waiting for is the page.click method, which should trigger the download.

scrapers/bank_of_malawi.py
gizmo572: So the download_info object is an instance of the Download class of Playwright, and should always contain the ‘value’ and ‘suggested_filename’ properties. 

The ‘suggested_filename’ property is based on the ‘filename’ property in the ‘content-disposition’ header of the HTTP response. Content-disposition defaults to “inline”, meaning content is loaded in the web browser, and when it is explicitly set to “attachment”, this indicates that the content is to be downloaded, in which case, a ‘filename’ property typically exists and is used to determine the name of the file. This is not always the case, but in the case of the Malawi scraper, it seems that this is the standard.

https://playwright.dev/docs/api/class-download#download-suggested-filename

scrapers/bank_of_malawi.py
nattiechan: So I gather you are thinking about this from a performance standpoint. However, I think 30 seconds for now is totally reasonable, especially we are scraping at most 10 circulars. I'd prefer not complicating things at this point

gizmo572: ok, I removed it.

nattiechan (approved):
Good work!

SFJohnson24 (approved):
looks good

PR #13, opened 2023-12-04
create scraper for Bank of Botswana.

gizmo572:
use bank_of_kenya.py as template for bank_of_botswana.py scaper.

The 'href' in the anchor tag that represents a banking circular link is used to obtain the url of the circular file.  If the file is not a pdf file, it is skipped so that the first pdf file is selected.

If no pdf file is found, an error is raised.


nattiechan:
So, I looked at the publication page for this bank, and it seems like this bank tends to publish monthly/annual financial statements in addition to other things. Before we review the code, when @sgoldfarb2 comes back, can we confirm whether we want to download the latest files available indiscriminately regardless of the content or if there are specific patterns that we can potentially look for and automate?

However, one caveat @sgoldfarb2 is that if what we need to filter out/identify content-wise is too random, we may not be able to implement it.

nattiechan:
Need help from @sgoldfarb2 regarding this: https://github.com/sgoldfarb2/scraper-infra/pull/13#issuecomment-1841283032

PR #15, opened 2023-12-06
Resolve #99: Decouple logic for downloading directory and file path

gizmo572:
scrapers/utils.py: 
-Rename “download_filepath” context manager to “download_directory” and pull logic for determining file path out of “download_directory”. Yield “directory” instead of “filepath”.
-Create “get_file_path” function for determining file path of a new file. 
-Adjust “get_standardized_filename” function to remove timestamp from the resulting file name, and replace it with a ‘cleaned’ version of the file’s parsed_url.

Adjust existing scrapers (bank_of_kenya.py and bank_of_zambia.py and bank_of_rwanda.py) per changes to utils.py: 
-replace references to “download_filepath” with “download_directory”
-import “get_standardized_filename” and “get_file_path” functions from utils.py
-invoke these functions to retrieve a files name and path immediately following directory creation.

bank_of_rwanda scraper:
-parse the 'link' variable to obtain a unique identifying 'Hash' and pass this to 'get_standardized_filename' instead of 'link'. This solves the error caused by the resulting file name being too long. 

create utils.test.py file in scrapers directory:
-add tests for 'get_standardized_filename' function of utils.py


scrapers/utils.py
```diff
@@ -59,13 +58,22 @@ def get_download_directory(filepath, keep_file):
...
+
+def get_standardized_filename(scraper_filepath, url):
     filepath_fragments = scraper_filepath.split("/")
     if not (filename := filepath_fragments[-1]).endswith(".py"):
         raise ValueError("Expected a python file.")
     bank_name = filename.replace(".py", "")
-    # TODO: Timezone for the dates?
-    return f"{filename}_{datetime.today().strftime('%Y%m%d')}.pdf"
+
+    chars_to_replace = "/ "
```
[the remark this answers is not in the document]
nattiechan: And if we have a number of characters we need to replace, I am starting to wonder if regex (i.e. `re`) would be an easier option regarding code readability.

I know that `maketrans` is faster than regex, but since we are not replacing a very long string, I wonder if the ease of readability may be more beneficial here.

scrapers/utils.py
```diff
@@ -59,13 +58,22 @@ def get_download_directory(filepath, keep_file):
     return directory
 
 
-def get_standardized_filename(scraper_filepath):
+def get_file_path(directory, filename):
+    file_path = f"{directory}/{filename}"
+    return file_path
+
+
+def get_standardized_filename(scraper_filepath, url):
```
nattiechan, line 66: To your point, `url` technically can be ID, the download URL, or something else entirely. The actual values being passed in are highly dependent on what is available on the website. I wonder if there is a more generalized name for the `url` argument

nattiechan: Also, can we do a check to ensure `scraper_filepath` and `url` are both `str` type and not empty string?

nattiechan (changes requested):
Looks great so far!

scrapers/utils.py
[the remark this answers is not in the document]
gizmo572: I did a bit of research and added some more characters that can be an issue in file names. Here is a source: https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file

scrapers/utils.py
```diff
@@ -60,13 +62,28 @@ def get_download_directory(filepath, keep_file):
     return directory
 
 
-def get_standardized_filename(scraper_filepath):
+def get_file_path(directory, filename):
+    file_path = f"{directory}/{filename}"
+    return file_path
+
+
+def get_standardized_filename(scraper_filepath: str, parsed_url: str):
```
nattiechan, line 70: I know Python supports typing, and my preference is to have typing as well. However, since our repo is not enforcing typing, let's not add typing to the arguments just to keep things consistent.

scrapers/utils.py
```diff
@@ -2,36 +2,38 @@
...
+    Context manager to create a directory for the scraper to store downloaded files.
     arg `scraper_filepath` should always be the filepath of the scraper file
     (i.e. `__file__`)
 
     Based on: https://stackoverflow.com/a/22726782
 
     Example:
-    with download_filepath(__file__) as filepath:
+    with download_directory(__file__) as directory:
+        filename = get_standardized_filename(__file__, id)
```
nattiechan, line 27: can we not assume this is going to be an id in the second argument

```suggestion
        filename = get_standardized_filename(__file__, link)
```

scrapers/utils.py
```diff
@@ -60,13 +62,28 @@ def get_download_directory(filepath, keep_file):
...
     if not (filename := filepath_fragments[-1]).endswith(".py"):
         raise ValueError("Expected a python file.")
     bank_name = filename.replace(".py", "")
-    # TODO: Timezone for the dates?
-    return f"{filename}_{datetime.today().strftime('%Y%m%d')}.pdf"
+
+    parsed_url = parsed_url.replace(".pdf", "")
+    clean_url = re.sub(r'[/\\:*?"<>|%! ]+', "_", parsed_url)
+
+    return f"{bank_name}_{clean_url}.pdf"
```
nattiechan, line 86: Few things:

1. I just realized that we assume all downloaded files are PDF. While that may be the case in practice, having that unspoken assumption here is not the best thing in the world. I think we should have an extra `kwarg` in this method:

```
# Let's make the second argument more generalized as well
def get_standardized_filename(scraper_filepath, url, file_extension='.pdf'):
   <the function>
```

So, individual scrapers can adjust the file extension as needed.

And then, instead of replacing `.pdf`, we will need to figure out where the file extension is towards the end of `url` argument, and replace the letters accordingly.

Then at the end, we return

```
return f"{bank_name}_{clean_url}{file_extension}"
```

2. Can we add the link that you used to research the symbols that we need to clean as a comment above L84 here?

scrapers/utils.test.py
```diff
@@ -0,0 +1,42 @@
+import unittest
+
+from utils import get_standardized_filename
+
+
+class TestUtilFunctions(unittest.TestCase):
+    # characters that should be replaced: ``` /\:*?"<>|%! ```
+    def test_get_standardized_filename(self):
```
nattiechan, line 8: In general, each test should only be testing one thing, or there is one operation and you are trying to assert multiple things from that one operation.

So with this one test you will need to break things up into their separate tests which is the best practicing for writing tests.

Last thing - make sure you test for `None` passed into `get_standardized_filename()` as discussed above. Perhaps it's worthwhile to add a test for different file extension passed into the function as well

Side bar - how do you expect contributors to run this test locally? Perhaps we can add some instructions above the class?

nattiechan (changes requested):

scrapers/utils.py
```diff
@@ -60,13 +62,28 @@ def get_download_directory(filepath, keep_file):
...
+
+
+def get_standardized_filename(scraper_filepath: str, parsed_url: str):
+    if not isinstance(scraper_filepath, str) or not isinstance(parsed_url, str):
+        raise TypeError("Both 'scraper_filepath' and 'parsed_url' must be strings.")
+    if scraper_filepath == "":
+        raise ValueError("Expected the string 'scraper_filepath' to have length.")
+    if parsed_url == "":
+        raise ValueError("Expected string 'parsed_url' to have length.")
+
```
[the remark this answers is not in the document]
gizmo572: Regarding point 1, I added tests for None and the isInstance method seems to be taking care of them as all tests are passing.

Regarding point 2, streamlining makes sense to me, but the potential issue I see is raising a ValueError instead of a TypeError when the error is caused by a non-string argument being passed in. Do you think we should raise a ValueError for all cases?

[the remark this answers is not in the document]
gizmo572: Point 1: I found this demo for the isInstance method. When you replace the string "Hello" with None, the appropriate boolean value is still returned:
https://www.w3schools.com/python/trypython.asp?filename=demo_ref_isinstance2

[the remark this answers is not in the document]
nattiechan: > Point 1: I found this demo for the isInstance method. When you replace the string "Hello" with None, the appropriate boolean value is still returned:

Cool - as long as it does what we intend them to do. That said, I think we still need the `not scraper_filepath` part to take care of `""` getting caught as invalid string

> Regarding point 2, streamlining makes sense to me, but the potential issue I see is raising a ValueError instead of a TypeError when the error is caused by a non-string argument being passed in. Do you think we should raise a ValueError for all cases?

It's a bit of a moot point tbh - I think ValueError captures the essence a bit better because at the end of the day, the `TypeError` is a sub-issue of not having the "right value" per see. Does that make sense?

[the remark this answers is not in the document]
gizmo572: This makes sense. My reservation stems from us being less specific about what exactly is causing the error.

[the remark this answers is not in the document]
nattiechan: Being less specific here is actually intentional - sometimes it would be just as helpful to have a more generalized message that captures the type of mistake then being overly specific. I would expect developers who see this error to check their inputs, which is the intention here anyway

scrapers/utils.py
```diff
@@ -68,12 +68,13 @@ def get_file_path(directory, filename):
...
 
 def get_standardized_filename(scraper_filepath: str, parsed_url: str):
-    if not isinstance(scraper_filepath, str) or not isinstance(parsed_url, str):
-        raise TypeError("Both 'scraper_filepath' and 'parsed_url' must be strings.")
-    if scraper_filepath == "":
-        raise ValueError("Expected the string 'scraper_filepath' to have length.")
-    if parsed_url == "":
-        raise ValueError("Expected string 'parsed_url' to have length.")
+    if (
+        not scraper_filepath.  # This takes care of checking for `None` and `""`
```
nattiechan, line 72: This comment is only for your info - could we take that out please and TY :pray:

scrapers/utils.test.py
```diff
@@ -0,0 +1,53 @@
+import unittest
+
+from utils import get_standardized_filename
+
+
+class test_get_standardized_filename(unittest.TestCase):
```
nattiechan, line 6: Let's add one more test with a different file extension passed in (`.docx`?) to make sure the file extension changed as a result

scrapers/utils.test.py
nattiechan, line 1: Can we add some instructions above the `class` in L6 regarding how to run this test? Is it just `python scrapers/utils.test.py`?

scrapers/utils.py
```diff
@@ -60,13 +62,31 @@ def get_download_directory(filepath, keep_file):
...
+
+
+def get_standardized_filename(scraper_filepath, parsed_url, file_extension=".pdf"):
+    if (
+        not scraper_filepath
+        or not isinstance(scraper_filepath, str)
+        or not parsed_url
+        or not isinstance(parsed_url, str)
+    ):
+        raise ValueError("Invalid arguments.")
```
nattiechan, line 77: I am debating between doing checks this way or use `assert`. Curious what your thoughts are @mlamchamkee

nattiechan (changes requested):
Very close! Just a few minor things

scrapers/utils.test.py
gizmo572: that's correct. I just added this.

nattiechan (approved):
Part of me still wonder if we should put the tests in a test folder, but I think you have done enough work with this that I will let the folks [here](https://github.com/sgoldfarb2/scraper-infra/pull/10#pullrequestreview-1776130832) to figure that out

gizmo572:
> Part of me still wonder if we should put the tests in a test folder, but I think you have done enough work with this that I will let the folks [here](https://github.com/sgoldfarb2/scraper-infra/pull/10#pullrequestreview-1776130832) to figure that out

I was actually going to do this originally, but ran into some complications. In hindsight, making it work isn't too complicated. This is what would be required at the top of the test file, assuming the test folder is located in the root directory:

```
import sys
import os

sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), '..', ‘scrapers’)))

import utils.test.py
```

nattiechan:
I was going to suggest putting the test folder within the `scrapers` folder - everything pertaining to scrapers needs to be contained within that folder for now. Does that change anything?

scrapers/utils.py
mlamchamkee: I think both are valid and it's a trade off between conciseness and robustness/safety.  While an `assert` statement has the benefit of being concise, they are omitted when the optimized option is turned on (`python -o main.py`)

https://discuss.python.org/t/best-practice-for-type-checking-and-assert-statement/33728

scrapers/utils.test.py
```diff
@@ -0,0 +1,61 @@
+import unittest
+
+from utils import get_standardized_filename
+
+# COMMAND TO RUN TESTS: ``` python scrapers/utils.test.py ```
+
+
+class test_get_standardized_filename(unittest.TestCase):
+    def test_replaces_chars(self):
```
mlamchamkee, line 9: Love to see a comprehensive suite of test cases for both valid and invalid scenarios!

mlamchamkee (dismissed):
Excellent job on this PR!  I left a couple of small questions, more so for educational purposes.

scrapers/utils.test.py
```diff
@@ -0,0 +1,61 @@
+import unittest
```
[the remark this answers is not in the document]
nattiechan: Honestly, we can consider it! I can whip up a spike ticket and see if anyone is keen to pick it up. Let's not block this from merging though (which I don't think you are)

scrapers/utils.py
```diff
@@ -60,13 +62,31 @@ def get_download_directory(filepath, keep_file):
...
 
-def get_standardized_filename(scraper_filepath):
+def get_file_path(directory, filename):
+    file_path = f"{directory}/{filename}"
+    return file_path
+
+
+def get_standardized_filename(scraper_filepath, parsed_url, file_extension=".pdf"):
+    if (
+        not scraper_filepath
```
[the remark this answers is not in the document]
nattiechan: >  Is this a validation to guard against no argument being passed in

It's more so passing in invalid values. If you look closely at the actual scrapers, the values that are being passed in originate from the elements on the DOM, which can be flakey at times. This is more so some defensive programming in action to protect against something that is inherently flakey (aka web scrapers)

scrapers/utils.py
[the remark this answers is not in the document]
nattiechan: Ya we talked about Type vs. ValueError earlier as well - I suggested `ValueError` because my intention is to protect against invalid values, which invalid type can be a case of invalid value if that makes sense.

scrapers/utils.py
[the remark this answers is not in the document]
mlamchamkee: I see!  I thought lines 72 and 74 were guarding against a developer not passing in the required number of required arguments when invoking the `get_standardized_filename` function.

scrapers/utils.test.py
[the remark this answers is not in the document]
nattiechan: https://github.com/sgoldfarb2/scraper/issues/111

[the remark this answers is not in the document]
mlamchamkee: We got burnt once because we realized we wanted it and it was a pain to convert an entire built out test suite from unittest to pytest

nattiechan:
@gizmo572 this should be good to go whenever you are ready to merge! 🎉

gizmo572:
> I was going to suggest putting the test folder within the `scrapers` folder - everything pertaining to scrapers needs to be contained within that folder for now. Does that change anything?

I just looked into this and it looks like we actually wouldn't have to modify 'sys.path' for this, so this would work as is

gizmo572:
> @gizmo572 this should be good to go whenever you are ready to merge! 🎉

awesome! I see that a scraper for namibia has been added since I last merged main into this branch, so I will first merge to main again and adjust the namibia scraper file

scrapers/utils.py
```diff
@@ -60,13 +62,31 @@ def get_download_directory(filepath, keep_file):
...
+        raise ValueError("Invalid arguments.")
+
     filepath_fragments = scraper_filepath.split("/")
     if not (filename := filepath_fragments[-1]).endswith(".py"):
         raise ValueError("Expected a python file.")
     bank_name = filename.replace(".py", "")
-    # TODO: Timezone for the dates?
-    return f"{filename}_{datetime.today().strftime('%Y%m%d')}.pdf"
+
+    parsed_url = parsed_url.replace(".pdf", "")
```
nattiechan, line 84:
```suggestion
    parsed_url = parsed_url.replace(file_extension, "")
```

nattiechan (changes requested):
There is one last thing that we should fix before merging.

A few things:

1. Since we have a [spike](https://github.com/sgoldfarb2/scraper/issues/111) looking into using a Python testing package, let's not worry about moving the test files for now. We will take care of that once we decide on Issue 111.
2. In general - when you push up a new commit it will dismiss the "stale" review from before requiring the reviewers to re-review your PR. We have a lot of parallel work going on along with this ticket, so I don't think it is realistic to merge new changes into `main` when there is a new scraper etc. just to keep this PR up to date. That will be quite a bit of burden on reviewers and blocking progress/other tickets that depend on this PR. So with that in mind, after this one more change and review let's just merge this in. We can always open another PR to address any remaining scrapers that need to be updated.

nattiechan (approved):
:tada:

nattiechan:
Sorry to ping you one more time @mlamchamkee let's give this a final review and get this merged in

mlamchamkee (approved):

PR #18, opened 2023-12-12
Re-scaffold project with projen, refactor CDK infrastructure code

jiedong111:
# Pull Request Overview

## Refactoring AWS CDK Code

This PR aims to refactor the existing AWS Cloud Development Kit (CDK) code to align with best practices, including:

- **File Structure Optimization:** Reorganize the file structure for intuitive navigation.
- **Stack Organization:** Separate stateful resources into their own stacks.
- **L3 Constructs Implementation:** Transition logical groupings/patterns from a collection of Level 2 (L2) constructs into Level 3 (L3) constructs for efficiency.

## Implementing Projen

- **Projen Introduction:** Projen is essentially AWS CDK for project scaffolding. It's highly recommended for CDK projects due to its seamless integration and development by core AWS CDK team members.
- **Benefits of Projen:** Projen offers sensible defaults for AWS CDK projects and allows for easy programmatic adjustments and distribution of repository settings as projects scale.
- **Further Reading:** Check out the [Projen Documentation](https://github.com/projen/projen) for more information.

## CDK Code Overview

- **DynamoDB Tables:** Defined in `uuidTableStack` and `ScraperTableStack` within `src/stacks`.
- **Data Population:** Initiated via `PopulatorStack` for initial data feeding into DynamoDB tables.
- **ScraperStack Functions:**
  - Definition of a topic, necessary queues, and instances of Scraper, Validator, and Notifier custom L3 constructs.
  - Execution of business logic through various Python scripts (`scraper.py`, `validator.py`, `stream-lambda.py`).

## Current Business Logic

1. **Scraper Functionality:** `scraper.py` simulates a scraper and constructs preliminary items for `scraperAggregationFifoQueue`.
2. **Validation Process:** `validator.py` validates hashes against a master list in `uuidTable`, updates the master list with new hashes, and places completed items in `scraperTable` after necessary transformations.
3. **Notification Events:** `stream-lambda.py` triggers notifications to `notificationFifoQueue` based on specific checks from `scraperTable`.

## Next Steps

- **Setting Up CI/CD:** It'll get harder to set up the longer we put it off, I think here is a good place to make this a focus
- **Docker Integration and Testing:** Look to set up docker images and test an actual scraper running in AWS
- **Minor Infrastructure Adjustments:** Address remaining minor issues in the infrastructure code.

Infra code is still a bit rough around the edges, but I think this is a good place for a PR.


README.md
nattiechan, line 1: For the intents and purposes of how other contributors will use this repo, I think we can keep most of this. If you don't want anyone to touch the AWS CDK then maybe take that section out from `README.md`

package.json
nattiechan, line 1: I assume a lot of these changes are generated by `projen`? and I guess we are using `yarn.lock` instead of `package-lock.json`?

scrapers-local/scrapers/Pipfile
nattiechan, line 1: Is there a reason why we are moving this to `/scrapers-local`? FYI that if we are using Docker Container down the road it's a major PITA to work with `Pipfile` that is not in the root folder

scrapers-local/scrapers/bank_of_kenya.py
nattiechan, line 1: Similarly - is there a reason why we are renaming the scraper folder? Technically these scrapers are not "local" if you know what I mean

src/constants.ts
```diff
@@ -0,0 +1,5 @@
+//space for holding necessary constants
+//in production this might contain multiple regions
+//may also contain information for multiple account environments
+export const PRIMARY_REGION = 'us-east-1';
```
nattiechan, line 4: Knowing that some bank circular pages cannot load with US IP address -Depending on where the primary users of this project would be (UK?) we may want to re-consider where the primary region should be

src/constructs/Notifier/Notifier.ts
```diff
@@ -0,0 +1,78 @@
...
+      handler: 'notification.main',
+      timeout: Duration.seconds(300),
+      //CDK may have pushed an update, it seems 3.12 is now supported?
+      //cdk-nag aspect also gives the follow error if you don't use latest runtime:
+      //[Error at /scraper-stack/scraper-notifier/notification-function-handler/Resource]
+      //AwsSolutions-L1: The non-container Lambda function is not configured to use the latest runtime versnion.
+      //Use the latest available runtime for the targeted language to avoid technical debt.
+      //Runtimes specific to a language or framework version are deprecated when the version reaches end of life.
+      //This rule only applies to non-container Lambda functions.
+      runtime: Runtime.PYTHON_3_12,
```
nattiechan, line 39: Are we using Python 3.11 or 12?

nattiechan (commented):
I have not looked closely at the AWS-specific code yet - I want to get the questions around overall architecture addressed first

src/constructs/Notifier/Notifier.ts
jiedong111: I think we originally went with 3.11 because CDK didn't support 3.12? Unless I'm remembering incorrectly. It seems to be generally against best practices for non-containerized Lambda functions based on the cdk-nag pack I was running to not run on the latest available runtime for that given language, but if we were using 3.11 for a specific reason then we can just ignore this I'm pretty sure.

src/constants.ts
jiedong111: Definitely something to think about yeah. We can deploy our primary region to wherever makes the most sense for our users.

package.json
jiedong111: We can configure projen to run with whatever package manager we feel like using. I'm not actually familiar with enough with the differences between like npm and yarn to have an informed opinion about this. Projen by default uses yarn though

.eslintrc.json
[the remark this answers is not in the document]
jiedong111: We can configure the linting according to whatever we need it to be by editing the projenrc.ts file if we want to edit the linting. Can look into figuring out how to mess with the projenrc to align with our current linting practices if necessary

scrapers-local/scrapers/Pipfile
jiedong111: I wasn't sure how it would interact with what projen scaffolded, so I wanted to keep them as separate as possible, can test moving `Pipfile` etc. back to root and see if anything weird happens. I think it's worth discussing moving either the infra or the scrapers to another repo though just to separate concerns as in other comment

src/constructs/Notifier/Notifier.ts
nattiechan: We can use Python 3.12 no problem, but that means we need to upgrade `Pipfile` to using 3.12 as well

src/constants.ts
nattiechan: @sgoldfarb2 for this question ^^^

.eslintrc.json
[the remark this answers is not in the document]
nattiechan: Or can we just audit and see what the differences are?

package.json
nattiechan: There are always endless debates about `npm` vs `yarn`. Personally I don't have a problem either way, though let's just make sure all our documentation aligns with whatever we are using

scrapers-local/scrapers/Pipfile
jiedong111: Moved `Pipfile`, `Pipfile.lock` and `pyprojec.toml` back to root and didn't seem to cause issues when running scrapers or `cdk watch`. I think it's still worth asking if we actually want the scrapers and the infrastructure in the same repo though

nattiechan: Is it a requirement to get the AWS infrastructure to work? If not, then I think keeping them separate may be helpful. The only thing is that if we want to use Docker images for both the CDK and the scrapers themselves we will need to maintain 2 separate images. Though the up side for that is that the Docker image for AWS will be substantially easier because we are running them in headless mode

PR #21, opened 2023-12-14
feature: added bank of mozambique scraper

SFJohnson24:
>added scrapper for bank of mozambique

>implemented similar logic to bank of malawi, allowing iteration through multiple circulars.  bank of mozambique does publish multiple circulars in a day historically so if we want to check for date, we will likely need to revisit this scraper and insert some logic in the loop to check for dates of circulars


scrapers/bank_of_mozambique.py
```diff
@@ -0,0 +1,51 @@
...
+                            file.write(response.content)
+                        md5hash = get_md5hash(filepath)
+                        
+                    if parse_arguments().debug:
+                        print(f"MD5hash for downloaded file: {md5hash}")
+                except:
+                    browser.close()
+                    raise   
+
+        browser.close()
```
nattiechan, line 47: So I like how you are trying to integrate some variations into the code here. That said, nested anything (nested for loops, context managers) adds complexity and I don't think we need the added complexity here at all.

```suggestion
        pdf_links_count = page.eval_on_selector_all(
            "div.download > a", "elements => elements.length"
        )
        count = min(n, pdf_links_count)
        link_array=page.locator("div.download > a").all()[:count]
        session = requests.Session()
        
        for link in link_array:
             download_link = link.get_attribute("href")[1:]
             # Even though there are more things we can optimize here
             # Since we do have plans to revisit and update all these scrapers anyway
             # let's leave this as-is for now
             try:
                 absolute_url = validate_and_get_download_link(BASE_URL, download_link)
                 response = session.get(absolute_url)
                 with download_directory(__file__) as directory:
                     filename = get_standardized_filename(__file__, link.split("Hash=")[-1])
                     filepath = get_file_path(directory, filename)
                     with open(filepath, "wb") as file:
                         file.write(response.content)
                     md5hash = get_md5hash(filepath)
                        
                 if parse_arguments().debug:
                     print(f"MD5hash for downloaded file: {md5hash}")
             except:
                 browser.close()
                 raise   

        browser.close()
```

nattiechan (changes requested):

nattiechan (approved):

PR #22, opened 2023-12-14
add functionality to Bank of Zambia scraper #100

gizmo572:
This PR adds functionality to the Bank of Zambia scraper:

-Support for downloading multiple files added. By default, the most recent 10 files are downloaded.
-Separate files that are part of the same circular are identified and merged as a single pdf file. A single md5hash is then generated based on this merged file.
-The scraper now returns a dictionary in which the keys are the respective file names of each downloaded PDF, and each value is an object containing the corresponding file path and md5hash of that file. 

Additions to utils.py:

-add_pdf_to_writer function: formats file content and adds it to the current PdfWriter instance.

Library Additions:

-PyPDF2 has been added to ‘Pipfile’ as a package.


scrapers/bank_of_zambia.py
```diff
@@ -1,42 +1,87 @@
...
 
-        absolute_url = f"{BASE_URL}{link}"
+            if len(filepaths) > 1:
+                with download_directory(__file__) as directory:
+                    zip_name = get_standardized_filename(__file__, link_id, ".zip")
+                    zip_path = get_file_path(directory, zip_name)
+                    zip_pdfs(zip_path, filepaths)
+                file_dictionary[zip_name] = {"filepath": zip_path, "md5hash": md5hash}
+            else:
+                file_dictionary[filename] = {"filepath": filepath, "md5hash": md5hash}
```
nattiechan, line 81: Few comments here:

Zipping the files will most likely be fairly standard across scrapers. Let's move the operation to `utils.py`
```
      # This should be in utils.py
      def some_function_name(filepaths, <whatever kwargs we need>):
            if len(filepaths) > 1:
                with download_directory(__file__) as directory:
                    # Let's get `zip_pdfs()` to get the standardized filename and filepath
                    # as well, and return the zipped filepath name
                    return zip_pdfs(__file__, link_id)
             return filepath
```

Then here we can

```suggestion
            output_filepath = some_function_name(**kwargs) if len(filepaths) > 1 else filepath
            filename = <get it some way....>
            file_dictionary[filename] = {"filepath": output_filepath, "md5hash": md5hash}
```

scrapers/utils.py
```diff
@@ -134,3 +137,31 @@ def validate_and_get_download_link(base_url, download_link, check_content_pdf=Tr
...
+    md5 = hashlib.md5()
+    for file in filepaths:
+        try:
+            md5.update(Path(file).read_bytes())
+        except IsADirectoryError:
+            pass
+    return md5.hexdigest()
+
+
+def zip_pdfs(zip_path, filepaths):
```
nattiechan, line 152: Can we add test for this (after it's been updated)? Bonus is adding test for `calculate_checksum()` as well.

You _may_ need to create some mock files for this.

scrapers/bank_of_zambia.py
```diff
@@ -1,42 +1,87 @@
...
+    zip_pdfs,
 )
 
 BASE_URL = "https://www.boz.zm/"
 CIRCULAR_URL = f"{BASE_URL}circulars.htm"
 
 
-def bank_of_zambia_scraper():
+def get_link_id(link):
+    parsed_link = link[:20]
```
nattiechan, line 21: curious why the first 20 characters?

scrapers/bank_of_zambia.py
```diff
@@ -1,42 +1,87 @@
...
+            link = pdf_link.get_attribute("href")
+            if not link or not link.endswith(".pdf"):
+                browser.close()
+                raise ValueError("Invalid file URL.")
+
+            link_id = get_link_id(link)
+
+            map_link_id_to_indices.setdefault(link_id, []).append(i)
+
+            map_idx_to_link[i] = link
```
nattiechan, line 50: Is there a way we can abstract this to a utility function that accepts a list of links and returns whatever it is required for the next step, regardless of which scraper uses the function?

Also, the naming convention for the variables may need to be clearer as I don't know what `map_link_id_to_indices` and `map_idx_to_link` is supposed to do

nattiechan (changes requested):
Great start. I think we need to think of us as trying to build utility function/infrastructure that could serve the majority of the scrapers, rather than the other way around.

I have not fully looked through the code yet but the comments below should serve as starting points

nattiechan:
Also - looks like your PR contains changes to `pipfile.lock`, which is NOT what we want :/ Unfortunately you did not separate the installation from code ([here](https://github.com/sgoldfarb2/scraper-infra/pull/22/commits/77b6b370338407a5a8b3b6cbbbfd795facc42b4c)) so it would be hard to revert.

Let's address the comments and clean things up first. Then at the end we will squash all the commits so we can take out the pipfile.lock changes.

scrapers/bank_of_zambia.py
gizmo572: 20 is an arbitrary number because we're comparing the dates to find files that are part of the same circular and some of the dates are formatted differently (i.e. 202301 vs. 2023-01 vs 20230114). So the length ends up being 14 to 17 characters, but it's a very small sample, so I used 20 just in case something like 2023-01-14 happens, which I guess would result in the beginning of the string being 19 characters total. 

The format of the href is "/Circular" followed by a numerical representation of the date (ex: 202301), followed by the title of the circular. Files from the same circular have the same date. There was never an instance of multiple different circulars having the same date, but I suppose this is possible if 2 different circulars were published in the same month. In hindsight, in the 2 instances where the circular includes 2 files, the entire href of the first file minus '.pdf' is included in the 2nd file, so I suppose I could check if the href of 1 file includes the entire href of another file (minus '.pdf') and vice versa. The problem is the sample size is so incredibly small that it's difficult to draw any definitive conclusions regarding syntax.

scrapers/bank_of_zambia.py
gizmo572: i thought about doing this, but 2 objects would need to be returned. We can return a tuple of objects and immediately destructure it. Would could also use a separate function for each object, but I think this would be a bit repetitive. I think this logic is definitely reuseable. 

Regarding the names, map_link_id_to_indices is a dictionary in which the keys are the dates extracted from the hrefs, and each value is a list of the indices (keys) at which each corresponding circular file is found in the map_idx_to_link object.  Maybe map_circular_to_files is better?

map_idx_to_link is a dictionary in which the key is the index at which the DOM element of a given file is found in the pdf_links list, and its value is its corresponding 'href' string. So maybe map_file_to_href is better

gizmo572:
> Also - looks like your PR contains changes to `pipfile.lock`, which is NOT what we want :/ Unfortunately you did not separate the installation from code ([here](https://github.com/sgoldfarb2/scraper-infra/pull/22/commits/77b6b370338407a5a8b3b6cbbbfd795facc42b4c)) so it would be hard to revert.
> 
> Let's address the comments and clean things up first. Then at the end we will squash all the commits so we can take out the pipfile.lock changes.

I see, I uninstalled everything on my local machine, but installations still persist on github.

nattiechan:
> > Also - looks like your PR contains changes to `pipfile.lock`, which is NOT what we want :/ Unfortunately you did not separate the installation from code ([here](https://github.com/sgoldfarb2/scraper-infra/pull/22/commits/77b6b370338407a5a8b3b6cbbbfd795facc42b4c)) so it would be hard to revert.
> > Let's address the comments and clean things up first. Then at the end we will squash all the commits so we can take out the pipfile.lock changes.
> 
> I see, I uninstalled everything on my local machine, but installations still persist on github.

Remember, uninstalling does not mean your `Pipfile.lock` will automatically restore to what it was before: https://discord.com/channels/1171867800726409316/1171869417756438619/1182744162328649831

scrapers/bank_of_zambia.py
nattiechan: Arbitrary number of anything is very fragile. We don't necessarily want to take that approach unless we have to.

How the scraper determine dates will be specific to each scraper. IMO this part we don't have to worry about generalizing things. As you mentioned that pattern looks to be as follows:
image

In this case I would probably use regex to use a pattern similar to `^/Circular[0-9]{4}-?[0-9]{2}` to get the string that we want. And if we cannot find something, make sure to raise an exception and fail the scraper

scrapers/bank_of_zambia.py
gizmo572: hey I pushed changes. I didn't see your comments here, but I did come to the same conclusion about not needing the 2nd object. 

I think the names will need to be changed (re: map not being a concept in python). 

Regarding your comment above about arbitrary numbers not being ideal, I actually refactored the code so the files are determined to be part of the same circular if the href of 1 file is found in the href of another file. The shorter file being what is used as the circular name (instead of the date).

gizmo572:
> > > Also - looks like your PR contains changes to `pipfile.lock`, which is NOT what we want :/ Unfortunately you did not separate the installation from code ([here](https://github.com/sgoldfarb2/scraper-infra/pull/22/commits/77b6b370338407a5a8b3b6cbbbfd795facc42b4c)) so it would be hard to revert.
> > > Let's address the comments and clean things up first. Then at the end we will squash all the commits so we can take out the pipfile.lock changes.
> > 
> > 
> > I see, I uninstalled everything on my local machine, but installations still persist on github.
> 
> Remember, uninstalling does not mean your `Pipfile.lock` will automatically restore to what it was before: https://discord.com/channels/1171867800726409316/1171869417756438619/1182744162328649831

so the commands I entered were "pipenv uninstall pypdf2", "pipenv lock".  Where pipenv lock is supposed to create a new Pipfile.lock based on the current Pipfile.

scrapers/bank_of_zambia.py
```diff
@@ -1,48 +1,70 @@
...
+                        filepaths.append(filepath)
+
+            md5hash = calculate_checksum(filepaths)
+            print("link_id", link_id, "md5hash", md5hash)
+
+            if len(filepaths) > 1:
+                zip_name, zip_path = zip_pdfs(__file__, filepaths, link_id)
+                file_dictionary[zip_name] = {"filepath": zip_path, "md5hash": md5hash}
+            else:
+                file_dictionary[filename] = {"filepath": filepath, "md5hash": md5hash}
```
nattiechan, line 58:
```suggestion
            output_filename = filename
            output_filepath = filepath
            if len(filepaths) > 1:
                zip_name, zip_path = zip_pdfs(__file__, filepaths, link_id)
                output_filename = zip_name
                output_filepath = zip_path
           file_dictionary[output_filename] = {"filepath": output_filepath, "md5hash": md5hash}
```

scrapers/bank_of_zambia.py
```diff
@@ -1,48 +1,70 @@
...
-            md5hash = get_md5hash(filepath)
+                with download_directory(__file__) as directory:
+                    filename = get_standardized_filename(__file__, curr_link)
+                    filepath = get_file_path(directory, filename)
+                    with open(filepath, "wb") as file:
+                        file.write(response.content)
+                        filepaths.append(filepath)
+
+            md5hash = calculate_checksum(filepaths)
+            print("link_id", link_id, "md5hash", md5hash)
```
nattiechan, line 52: Let's remove these print statements for debugging purposes

scrapers/bank_of_zambia.py
```diff
@@ -1,48 +1,70 @@
...
+                file_dictionary[zip_name] = {"filepath": zip_path, "md5hash": md5hash}
+            else:
+                file_dictionary[filename] = {"filepath": filepath, "md5hash": md5hash}
 
         if parse_arguments().debug:
             print(f"MD5hash for downloaded file: {md5hash}")
 
         browser.close()
 
+        print("file_dictionary", file_dictionary)
```
nattiechan, line 65: same here

nattiechan (changes requested):
Once we clean this up, I think we are ready to add more tests for the utility functions

scrapers/utils.py
```diff
@@ -134,3 +137,71 @@ def validate_and_get_download_link(base_url, download_link, check_content_pdf=Tr
...
+    map_circular_to_links = {}
+    circulars = set()
+
+    for pdf_link in pdf_links:
+        link = pdf_link.get_attribute("href")
+        if not link or not link.endswith(".pdf"):
+            browser.close()
+            raise ValueError("Invalid file URL.")
+
+        file_name = link.replace(".pdf", "")
```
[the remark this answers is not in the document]
gizmo572: This is actually a point of confusion for me right now. Do we want to account for the possibility of non-pdf files for every Bank, or just for Banks which have a history of publishing circulars that are not pdf files? If we want to account for non-pdf files, then the error handling above on line 182 needs to be changed to remove the '.pdf' check.

scrapers/utils.py
```diff
@@ -134,3 +137,71 @@ def validate_and_get_download_link(base_url, download_link, check_content_pdf=Tr
...
+                circulars.add(file_name)
+                map_circular_to_links[file_name] = map_circular_to_links[circular]
+                map_circular_to_links[file_name].append(link)
+                del map_circular_to_links[circular]
+                file_added = True
+                break
+
+        if not file_added:
+            circulars.add(file_name)
+            map_circular_to_links[file_name] = [link]
```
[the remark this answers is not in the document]
gizmo572: The reason this is convoluted is because we are populating the 'map_circular_to_links' object as we are iterating through the files. So we are accounting for the possibility that we see the 'attachment' file before the main file, in which case the attachment file is initially added to the dictionary, then removed once it becomes clear that it is an attachment.

I agree that this utility function is probably not very reusable in this state. The problem I see is that in determining which files are a part of the same circular, it is not much extra to just build the 'map_circular_to_links' object right there instead of using a utility function to do so. I will think more about how to potentially make this logic reusable.

scrapers/utils.py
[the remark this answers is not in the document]
nattiechan: We can always have a `file_extension=".pdf"` `kwarg` in the function to handle that situation

scrapers/utils.py
```diff
@@ -134,3 +137,71 @@ def validate_and_get_download_link(base_url, download_link, check_content_pdf=Tr
...
+        file_name = link.replace(".pdf", "")
+
+        matching_file = None
+
+        # iterate through keys added thus far to the dictionary
+        for circular in circular_to_links_dict.keys():
+            if circular in file_name or file_name in circular:
+                # if a matching file is found, reassign 'matching_file' to that file name
+                matching_file = circular
+                break
```
nattiechan, line 193: I _think_ you can get the equivalent result using `re.search()` instead and it's easier to understand: https://www.geeksforgeeks.org/python-substring-key-match-in-dictionary/ (method 4)

```
matching_circulars = [circular for circular in circular_to_links_dict if re.search(file_name, circular)]
if matching_circulars:
    # May want to do a check to ensure we only have 1 item in the list
    matching_file = matching_circulars[0]
```

scrapers/bank_of_zambia.py
```diff
@@ -1,48 +1,71 @@
...
+
+            md5hash = calculate_checksum(filepaths)
+
+            output_filename = filename
+            output_filepath = filepath
+            if len(filepaths) > 1:
+                zip_name, zip_path = zip_pdfs(__file__, filepaths, link_id)
+                output_filename = zip_name
+                output_filepath = zip_path
+            file_dictionary[output_filename] = {"filepath": output_filepath, "md5hash": md5hash}
```
nattiechan, line 59: I think we can safely move this into a utility function, because this will most likely resulting in duplicated code across all scrapers anyway

scrapers/utils.py
```diff
@@ -134,3 +137,71 @@ def validate_and_get_download_link(base_url, download_link, check_content_pdf=Tr
...
+                        print(f"File '{file}' does not exist.")
+                except Exception as e:
+                    print(f"Error deleting file '{file}': {e}")
+    except Exception as e:
+        print(f"Error creating zip file: {e}")
+
+    return zip_name, zip_path
+
+
+def connect_circulars_to_associated_files(pdf_links):
```
nattiechan, line 176: I think we need a docstring explaining this function. It's not the easiest to understand and I don't know how much simpler we can get this

scrapers/utils.py
```diff
@@ -134,3 +137,71 @@ def validate_and_get_download_link(base_url, download_link, check_content_pdf=Tr
...
+                break
+
+        if matching_file:
+            # of the matching files, the 1 with the shortest length is used as the circular name
+            circular_name = min(file_name, matching_file)
+            # if the current file is shorter than the previously added file, replace the previous with the current in the dictionary
+            if circular_name == file_name:
+                circular_to_links_dict[file_name] = circular_to_links_dict.pop(matching_file)
+
+            circular_to_links_dict[circular_name].append(link)
```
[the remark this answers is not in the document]
gizmo572: Say we have a list of 2 hrefs:

Hrefs=[ '/test1_attachment.pdf',  '/test1.pdf' ]

In this case, /test1_attachment will initially be added to the dictionary because it comes first. When we get to '/test1.pdf', we see that it is part of the same circular as /test1_attachment.pdf, and it's shorter in length, so we want to replace the /test1_attachment key with a /test1 key.

If the attachment file comes after the main test1 file, we will just add the attachment file to the list that is a value of the test1 key in the dictionary.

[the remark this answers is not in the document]
gizmo572: It might make it easier to first iterate through the list of DOM elements and create a new list of hrefs with the extensions removed. Then we would have access to all file names right away and we could sort that list by length, guaranteeing that for circulars with multiple files, the shortest file names come first.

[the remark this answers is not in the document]
nattiechan: okay gotcha. Yeah I think if we just iteration through the list and sort the DOM elements first, then we can systematically sort each list with the shortest name comes first. Although it may like a touch less efficient it may be easier to understand and more generalizable

scrapers/bank_of_zambia.py
```diff
@@ -1,48 +1,85 @@
...
 
-        pdf_link = page.locator("div.list-group > a").all()[0]
-        link = pdf_link.get_attribute("href")
+        links = page.locator("div.list-group > a").all()
+        count = min(n, len(links))
 
-        if not link or not link.endswith(".pdf"):
-            browser.close()
-            raise ValueError("Invalid file URL.")
+        file_names_and_exts = get_list_of_file_names_and_extensions(links[6:9])
```
nattiechan, line 48: Is there a reason why we use `links[6:9]` here specifically?

scrapers/utils.py
```diff
@@ -134,3 +137,77 @@ def validate_and_get_download_link(base_url, download_link, check_content_pdf=Tr
...
+
+        # iterate through the files added thus far to the dictionary
+        for circular in file_names[:i]:
+            # if the current file name is an extension of a file already added as a key to the dictionary, reassign the key to be added and break the loop
+            if circular[0] in curr_file:
+                key_to_be_added = circular[0]
+                break
+
+        # add the current file to the list of the 'key_to_be_added' key in the dictionary
+        circular_to_links_dict.setdefault(key_to_be_added, []).append(f"{curr_file}.{extension}")
```
nattiechan, line 199: Trying to loop through part of the iterator that we were already trying to loop through in the outer loop is not the best practice. We already sort the file names in L185, correct? That means we can look through the keys of the dictionary to find the key that we want to add to, correct?

```suggestion
    for i, [curr_file, extension] in enumerate(file_names):
        keys = [key for key in circular_to_links_dict if re.search(key, curr_file)]
        if (key_length := len(keys)) > 1:
            raise ValueError(f"Expected 1 matched key for the file, found {key_length}")
        key_to_be_added = keys[0] if keys else curr_file

        # add the current file to the list of the 'key_to_be_added' key in the dictionary
        circular_to_links_dict.setdefault(key_to_be_added, []).append(f"{curr_file}.{extension}")
```

scrapers/utils.py
```diff
@@ -134,3 +137,77 @@ def validate_and_get_download_link(base_url, download_link, check_content_pdf=Tr
...
+                    print(f"Error deleting file '{file}': {e}")
+    except Exception as e:
+        print(f"Error creating zip file: {e}")
+
+    return zip_name, zip_path
+
+
+def connect_circulars_to_associated_files(file_names):
+    """
+    Accepts 'file_names' as an argument, which is a list of lists in which each sub-list consists of a file name at index 0 and a file extension at index 1.
```
nattiechan, line 178: Let's include an example of how the actual `file_names` could look like

scrapers/utils.py
```diff
@@ -134,3 +137,77 @@ def validate_and_get_download_link(base_url, download_link, check_content_pdf=Tr
...
+                filename = os.path.basename(file)
+                zipf.write(file, arcname=filename)
+            for file in filepaths:
+                try:
+                    if os.path.exists(file):
+                        os.remove(file)
+                    else:
+                        print(f"File '{file}' does not exist.")
+                except Exception as e:
+                    print(f"Error deleting file '{file}': {e}")
```
nattiechan, line 169: Having 2 for loops iterating through the same iterable seems odd. Also, is there a real need for us to delete the files?

scrapers/utils.py
```diff
@@ -134,3 +137,77 @@ def validate_and_get_download_link(base_url, download_link, check_content_pdf=Tr
...
+    except Exception as e:
+        print(f"Error creating zip file: {e}")
```
nattiechan, line 171: If there is an exception, the way we write this right now the exception will never be raised and the scraper operation continues. Is that what we want?

scrapers/bank_of_zambia.py
```diff
@@ -1,48 +1,85 @@
...
+                absolute_url = f"{BASE_URL}{curr_link}"
+
+                response = session.get(absolute_url)
+
+                with download_directory(__file__) as directory:
+                    filename = get_standardized_filename(__file__, curr_link)
+                    filepath = get_file_path(directory, filename)
+                    with open(filepath, "wb") as file:
+                        file.write(response.content)
+                        filepaths.append(filepath)
```
nattiechan, line 67: I think we can put L62-66 into a util function. L67 should be able to operate outside of the context manager.

```
# in utils.py

def download_file(scraper_name, link):
   with download_directory(scraper_name) as directory:
      filename = get_standardized_filename(scraper_name, link)
      filepath = get_file_path(directory, filename)
         file.write(response.content)
    return file

# in bank_of_zambia.py

filepath = download_file(__file__, curr_link)
filepaths.append(filepath)
```

nattiechan (changes requested):
We are starting to get close! After this round of changes I think we are ready to add some more tests to the utility functions that we added in this PR

scrapers/bank_of_zambia.py
gizmo572: I forgot to change this. This is to speed up testing because it accounts for 1 circular consisting of 1 file and 1 circular consisting of 2 files.

scrapers/utils.py
gizmo572: This makes sense that we shouldn't use the same iterator twice in a nested loop. I like this change, but I wonder if it makes sense to retrieve a list of keys rather than just 1 key because the current logic makes it impossible for there to be more than 1 matching key in the dictionary.

scrapers/utils.py
gizmo572: So the zipf.write method is making copies of existing files and adding them to the zip file, rather than placing the existing files in a zip file. I can look into zipping them in place. Regarding the 2nd loop, I think my idea was to make sure the writing operation is finished before deleting files, but in hindsight I don't think that really adds value, and it would be fine to just delete each file in the first loop, right after writing it to the zip.

scrapers/bank_of_zambia.py
nattiechan: Bump on removing this debugging statement

scrapers/utils.py
nattiechan: Okay, having this discussion here made me realized I made a mistake in terms of how we are managing context managers between the zipped and non-zipped files.

When we use the context manager for `download_directory()` [here](https://github.com/sgoldfarb2/scraper-infra/blob/6ab5801f64ee87c378751e5486dd657d74677535/scrapers/bank_of_zambia.py#L62C17-L62C64), after we append the filepath (L67 in the scraper file) if we do not keep the files (which we won't by default) we will delete the directory [here](https://github.com/sgoldfarb2/scraper-infra/blob/6ab5801f64ee87c378751e5486dd657d74677535/scrapers/utils.py#L41), which will cause problems down the road.

What we need to do is:

- We need to first have a context manager (No.1) that creates the overall folder, for both zipped and non-zipped files
- When we need to create the folder for the zipped files, we need to create the folder WITHIN the directory of the overall folder (context manager No.2)
- We need to do all operations (zipping the folder, sending the file to s3... etc) with the top context manager (aka No.1) still "open", so when the manager closes it can delete the directory and everything that is inside.

You will need to fact check me on this, though I think we need to:

1. Update `get_download_directory()` to support taking a directory as an optional argument, so we can create a folder within the overall folder.

2. Here in `zip_pdfs()`:
```
def zip_pdfs(directory, scraper_filepath, filepaths, link_id):
    with download_directory(directory, scraper_filepath) as zipped_directory:
        zip_name = get_standardized_filename(scraper_filepath, link_id, ".zip")
        zip_path = get_file_path(zipped_directory, zip_name)

        with zipfile.ZipFile(zip_path, "w") as zipf:
            for file in filepaths:
                filename = os.path.basename(file)
                zipf.write(file, arcname=filename)
```

3. In the scraper file:
```
# Excuse the incorrect indentation - you get the point
with download_directory(__file__) as directory:
    for circular_name in circular_to_links_dict:
            filepaths = []

            for curr_link in circular_to_links_dict[circular_name]:
                absolute_url = f"{BASE_URL}{curr_link}"

                filename = get_standardized_filename(__file__, curr_link)
                filepath = get_file_path(directory, filename)
                   file.write(response.content)
                   filepaths.append(filepath)

            md5hash = calculate_checksum(filepaths)

            populate_file_dictionary(
                __file__, file_dictionary, filepaths, circular_name, filename, filepath, md5hash
            )

        if parse_arguments().debug:

        # We will need to upload files to s3 here before we close the context manager, when we get there
        browser.close()

return file_dictionary
```

@gizmo572 sorry to lead you down a rabbit hole. Let me know if you have questions.

scrapers/bank_of_zambia.py
nattiechan: UPDATE: See https://github.com/sgoldfarb2/scraper-infra/pull/22/files#r1433244538

scrapers/utils.py
gizmo572: ah, this makes sense: we need everything to happen within the execution context of a single context manager invocation, or else the files will be deleted prematurely in a production environment. I will make the changes.

scrapers/utils.py
```diff
@@ -53,9 +56,14 @@ def get_browser(playwright):
...
-def get_download_directory(filepath, keep_file):
+def get_download_directory(filepath, keep_file, root_directory=None):
     if keep_file:
-        directory = f"{Path(filepath).parent}/circular_files"
+        directory = (
+            f"{root_directory}/zip_files"
+            if root_directory
+            else f"{Path(filepath).parent}/circular_files"
+        )
+
```
nattiechan, line 66: we need to have the ability to create using root directory regardless of whether `keep_file` is true or not

```
def get_download_directory(filepath, keep_file, root_directory=None):
    if root_directory:
        # We should not allow duplicated folders here
        # and we will assume `root_directory` is an absolute path - will need to fact check here
        Path(f"{root_directory}/{filepath}").mkdir()
    elif keep_file:
        directory = f"{Path(filepath).parent}/circular_files"
        ......
```

Then in the scraper file:

```
with download_directory(circular_name, directory) as zip_directory:
  ......
```

scrapers/utils.py
```diff
@@ -14,7 +17,7 @@
 
 
 @contextmanager
-def download_directory(scraper_filepath):
+def download_directory(scraper_filepath, root_directory=None):
```
nattiechan, line 20: Let's be a bit more explicit (and always loosen the scope a bit later):

```suggestion
def download_directory(scraper_filepath, zip_root_directory=None):
```

scrapers/utils.py
```diff
@@ -134,3 +142,89 @@ def validate_and_get_download_link(base_url, download_link, check_content_pdf=Tr
...
+
+def connect_circulars_to_associated_files(file_names):
+    """
+    Accepts 'file_names' as an argument, which is a list of lists in which each sub-list consists of a file name at index 0 and a file extension at index 1.
+
+    This function identifies files that are part of the same circular and groups them together in a returned dictionary, in which each key is a circular name and its value is a list of all files belonging to that circular.
+
+    Example of valid input:     [ ["file1", "pdf"], ["file2", "pdf"], ["file3", "docx"] ]   OR    [ ("file1", "pdf") ]
+
+    Invalid input:      [ ["file1", 1] ]    OR    [ {"file1", "pdf"}, {"file2", "pdf"} ]    OR  { "file1": "pdf", "file2": "pdf" }
```
nattiechan, line 177: Probably don't need the invalid input - the common practice is for the devs to go into the test files to figure out the invalid input. We should encourage skill development here as well

scrapers/utils.py
```diff
@@ -134,3 +142,89 @@ def validate_and_get_download_link(base_url, download_link, check_content_pdf=Tr
...
+    """
+    circular_to_links_dict = {}
+    try:
+        # sort file_names list by file name string length
+        file_names.sort(key=lambda x: len(x[0]))
+    except Exception as e:
+        raise TypeError(f"Invalid 'file_names' argument.")
```
nattiechan, line 184: I would prefer letting `sort` throw the actual errors here. This way we are not masking the actual errors, making debugging difficult

scrapers/utils.py
```diff
@@ -134,3 +142,89 @@ def validate_and_get_download_link(base_url, download_link, check_content_pdf=Tr
...
+        try:
+            curr_file, extension = file_list
+            if (
+                not isinstance(curr_file, str)
+                or not isinstance(extension, str)
+                or (not isinstance(file_list, list) and not isinstance(file_list, tuple))
+            ):
+                raise
+        except Exception:
+            raise ValueError(f"The element at index {i} is invalid.")
```
nattiechan, line 196: That is not the correct use of try/except block. Also, if `file_list` is not in the format that we expect the way we won't be able to unpack the values properly anyway. I would not worry about that.

The moral of the story here is that trying to raise custom exception every time is not the correct approach. Sometimes allowing the operation to raise their own exception will facilitate easier debugging - not always though! So it's a bit of a trial and error and experience to understand what is what.

```suggestion
            curr_file, extension = file_list
            if (
                not isinstance(curr_file, str)
                or not isinstance(extension, str)
            ):
                raise TypeError(f"Incorrect type for file name and extension.")
```

nattiechan (changes requested):
This is starting to look really, really good!!!! Nice job so far! My comments are mostly cleaning up the util functions to make it production ready :)

One note regarding squashing commits -

Basically we need to squash all 40+ commits into one, so we can remove offending `Pipfile.lock` before force pushing the commit back here. I want you to take a shot at this, though **I highly recommend you create a separate branch to do a dry run first before executing the steps on the actual branch.**

1. Make sure you are on this branch `git switch updateZambia`
2. Create a new branch from this branch `git switch -c updateZambia2`
3. `git reset --soft HEAD~40` (the number 40 may change if you have additional commits - your goal is to reset all commits that you have done in the branch so far)
4. Remove the changes in `Pipfile.lock` (maybe easiest to do it in the VS code?)
5. Re-commit the changes, preferably with a descriptive commit message on what has been done
6. Look at the commit history via `git log` to see if those are the desired changes
7. If so, go back to the actual branch `git switch updateZambia` and repeat step 3-5
8. `git push -f` (yes, you need to force push to rewrite history)

Source: https://stackoverflow.com/a/5201642

Side note: You will get devs out there who absolutely despise force push and I get that. But here I think not having changes in `Pipfile.lock` is more important because of the circumstances that we are in.

scrapers/bank_of_zambia.py
```diff
@@ -1,48 +1,97 @@
...
+                        filepath,
+                        md5hash,
+                    )
 
         if parse_arguments().debug:
             print(f"MD5hash for downloaded file: {md5hash}")
 
         browser.close()
 
+        # print("file_dictionary", file_dictionary)
```
nattiechan, line 92: Remove the comment here please

scrapers/utils.py
gizmo572: My thought was that all zipped files would be stored in the same "zip-directory", and the reason for storing them separately from the other files is that we will ignore the regular ".pdf" files in the zip-directory and only upload the packaged ".zip" files to the S3 bucket. So something like this:
```
def get_download_directory(filepath, keep_file, root_directory=None):
    if root_directory:
        directory = f"{root_directory}/zip_files"
    elif keep_file:
        directory = f"{Path(filepath).parent}/circular_files"
    else:
        directory = tempfile.mkdtemp()

    Path(directory).mkdir(parents=True, exist_ok=True)

    return directory
```

nattiechan: > My thought was that all zipped files would be stored in the same "zip-directory", and the reason for storing them separately from the other files is that we will ignore the regular ".pdf" files in the zip-directory

I may be confused - how would we handle a situation where we need to download 10 circulars, but only 1 out of the 10 circulars need to be zipped?

scrapers/utils.py
gizmo572: I'm confused about this because the "root_directory" represents the root directory for the zip folder as well as all other circular files in this context.

scrapers/utils.py
gizmo572: This makes sense. I had the feeling that the default error handling might be more beneficial here due to the specificity.

scrapers/utils.py
gizmo572: if we download 10 circulars and 1 of them needs to be zipped, that would mean we have at least 11 files because the circular that needs to be zipped consists of at least 2 files. Assuming it is 2 files, the root directory would consist of 9 pdf files and 1 zip folder. Inside the zip folder would be 2 pdf files and 1 zip file that contains those 2 pdf files. My assumption was that we want to upload the 9 pdf files in the root folder, as well as the 1 zip file in the zip folder, while ignoring the 2 pdf files in the zip folder.

nattiechan: Okay, so my biggest concern here is the assuming that when someone passes `root_directory` into the function, that they will for sure want `zip_file` to be the folder name. While it is true for the current use-case, since this is a utility function there is no stopping others from utilizing this function for other purposes.

Perhaps we can do something like

```
def get_download_directory(root_directory=None, filepath="circular_files", keep_file=False):
    if root_directory:
        # We will need to explicitly pass filepath="zip_files" here if desired
        directory = f"{root_directory}/{filepath}"
    elif keep_file:
        root_directory = root_directory if root_directory else Path(filepath).parent
        directory = f"{root_directory}/{flepath}"
    else:
        directory = tempfile.mkdtemp()

    # I think we need to make sure in the case where `directory = tempfile.mkdtemp()`,
    # this would not interfere with directory's ability to remove the directory on its own
    Path(directory).mkdir(parents=True, exist_ok=True)

    return directory
```

scrapers/utils.py
nattiechan: We can keep this as root_directory for now

scrapers/utils.py
gizmo572: It makes sense to me that we would want to pass in "zip_files" rather than explicitly stating it so that the function is more reusable. But I think we still need the actual 'filepath' variable as it currently exists for determining the 'root_directory' when it doesn't exist yet. Maybe we can add another parameter, 'folder_name' that defaults to 'circular_files', and that we will change to be 'zip_files' for the inner loop. 

```
def get_download_directory(filepath, keep_file, folder_name='circular_files', root_directory=None):
    if root_directory:
        directory = f"{root_directory}/{folder_name}"
    elif keep_file:
        directory = f"{Path(filepath).parent}/{folder_name}"
    else:
        directory = tempfile.mkdtemp()

    Path(directory).mkdir(parents=True, exist_ok=True)

    return directory
```

gizmo572: Regarding the case in which root_directory is None, I was under the impression that we could use the same logic when keep_file is True or False because we will eventually delete the root_directory, therefore deleting all sub-directories inside it as well. If we create temporary sub-directories for the zip files, then we will end up creating a unique folder for each zip file, which I suppose is not necessarily a bad thing, but it might complicate things if we want to have sub-directories for things other than zip files.

nattiechan: > But I think we still need the actual 'filepath' variable as it currently exists for determining the 'root_directory' when it doesn't exist yet

Good point. though I think we could be more explicit because of the way that we handle `filepath`

```
# Making all arguments keyword arguments here are somewhat intentional
# So we are more explicit about what the assumed defaults are for ease of understanding
def get_download_directory(
    root_directory=None,
    scraper_filepath=None,
    folder_name='circular_files',
    keep_file=False
):
    if root_directory:
        directory = f"{root_directory}/{folder_name}"
    elif keep_file:
        directory = f"{Path(scraper_filepath).parent}/{folder_name}"
    else:
        directory = tempfile.mkdtemp()

    Path(directory).mkdir(parents=True, exist_ok=True)

    return directory
```

> If we create temporary sub-directories for the zip files, then we will end up creating a unique folder for each zip file, which I suppose is not necessarily a bad thing, but it might complicate things if we want to have sub-directories for things other than zip files.

That's up to how the contributor decides to implement things, correct? Yes the utility function can provide some hints on how things should be done, but we cannot assume contributors are going to do things the way we want 100%. If they decide to go wild and create temp sub-directories, then it's on them to figure out how to manage all the sub-directories.

Plus, this is why code review exists to catch these novel/more complicated use-cases and decide if they are justified.

gizmo572: Regarding the use of explicit default parameters for ease of understanding, this makes sense to me. This will require us to adjust the way that 'get_download_directory' is invoked, but I don't think it's ever explicitly invoked in a scraper file right now, so that should be a quick fix. Are you ok with the version of this function that you have pasted here?

nattiechan: Let's do this to minimize changes for now:

```
def get_download_directory(
    scraper_filepath=None,    # Note the change in ordering
    root_directory=None,
    folder_name='circular_files',
    keep_file=False
):
    if root_directory:
        directory = f"{root_directory}/{folder_name}"
    elif keep_file:
        directory = f"{Path(scraper_filepath).parent}/{folder_name}"
    else:
        directory = tempfile.mkdtemp()

    Path(directory).mkdir(parents=True, exist_ok=True)

    return directory
```
Then all we need to do is:

Update `download_directory()` to:

```
def download_directory(scraper_filepath=None, root_directory=None, folder_name='circular_files'):
     .......
     directory = get_download_directory(scraper_filepath, root_directory, folder_name, inputs.keep_file)
     ......
```

Then, when we need to instantiate a zip directory

```
with download_directory(root_directory=directory, folder_name="zip_file") as zip_directory:
    .....
```

That gives us what we want while minimizing the impact for now. We can always come back and refactor if we want

scrapers/bank_of_zambia.py
```diff
@@ -1,48 +1,98 @@
...
 
-        pdf_link = page.locator("div.list-group > a").all()[0]
-        link = pdf_link.get_attribute("href")
+        links = page.locator("div.list-group > a").all()
+        count = min(n, len(links))
 
-        if not link or not link.endswith(".pdf"):
-            browser.close()
-            raise ValueError("Invalid file URL.")
+        file_names_and_exts = get_list_of_file_names_and_extensions(links[6:9])
```
nattiechan, line 48: Make sure we fix this part to account for `count`

```suggestion
        file_names_and_exts = get_list_of_file_names_and_extensions(links[:count])
```

nattiechan (commented):
Nice work! Now let's get a 2nd reviewer in here

Pipfile.lock
nattiechan, line 1: and let's make sure we clean this up asap now as well

PR #24, opened 2023-12-20
Jiedong111/scraper re

jiedong111:
# Infrastructure Update

## Local Setup Instructions
- Install CDK and bootstrap it for your AWS account
- Provide your email as `EMAIL=<your-email>` in a `.env` file in root of project
- Run `cdk deploy uuid-table-stack scraper-table-stack`
- Run `npx projen populate` and `npx projen retrieve`
- Run `cdk deploy scraper-stack`
- Check your email for a subscription notification and accept subscription

### Important Note
- The current way the stacks are made will send you 10 notifications a minute, and also run 10 lambda functions per minute. After verifying that the workflow is working as expected, it's best to tear down `uuid-table-stack`, `scraper-table-stack`, and `scraper-stack` so you don't accrue charges. You can do this by running `cdk destroy <stack-you-want-to-destroy>` or `cdk destroy --all`.

## Code Refactoring
- Abstracted out hackey `PopulatorStack.ts` functionality into the scripts in `src/taskScripts` that projen can run as tasks. Check out `.projen/tasks.json` if you want to get an idea for what that's looking like. The script `src/taskScripts/populateTable.ts` runs initial population of our `bank-uuid` and `scraper-run` tables. `src/getUUIDData.ts` is used to retrieve the relevant data for `src/stacks/ScraperStack.ts` to run its factory of scrapers. 

## Scraper Implementation
- I think the most salient points to go over are likely in `ScraperStack.ts`, are we comfortable with the data we're passing to each Scraper construct, does this seem like a scalable approach to store the metadata related to the provisioning of each scraper in our `bank-uuid-table`. Then it probably makes the most sense to look at each of the `scraper.py` functions in `src/scrapers` to check if how our object is being constructed aligns with what we're thinking, and if the orchestration of typing each of the scraper functions to it's own cron event rule is something that we're comfortable with moving forward. It's working noting right now that the scraper file reference in `bank-uuid-table` is just `scraper[i].main`, in production with how we're containerizing things this would either be a reference to an ECR image or an S3 bucket file where we're storing our scrapers.

### Next Steps
- I think if we're fine with how this workflow is looking, we can move on to containerizing our lambda functions to run an actual scraper in AWS.


src/taskScripts/populateTable.ts
```diff
@@ -0,0 +1,100 @@
...
+      master_hash_list: [{ hash_content: 'initial hash' }],
+      scraper_file: 'scraper3.main',
+      cron_rule: 'cron(* * * * ? *)',
+    },
+    {
+      uuid: uuidv4(),
+      bank_name: 'Bank of Kenya',
+      master_hash_list: [{ hash_content: 'initial hash' }],
+      scraper_file: 'scraper4.main',
+      cron_rule: 'cron(* * * * ? *)',
```
nattiechan, line 55: This is every minute?

nattiechan (commented):
IDK - this is surprisingly much easier to understand than the last iteration? good job!

nattiechan:
I would also like the build failure to get fixed as well. Since this is based on the `projen` branch, depending on how we want to execute the repo split/re-org we may want to merge the PR into `projen` so we can view the overall changes as well

src/.DS_Store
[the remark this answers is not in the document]
jiedong111: I am actually not sure what this file is, since going to my branch I actually don't see it represented, will double check one what `.DS_store` is doing

src/main.ts
```diff
@@ -1,6 +1,12 @@
-import { App } from 'aws-cdk-lib';
+import * as fs from 'fs';
+import path from 'path';
+import {
+  App,
+  // Aspects,
+} from 'aws-cdk-lib';
 // import * as nag from 'cdk-nag';
```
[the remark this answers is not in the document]
jiedong111: I left those in mostly for ease-of-use in terms of testing deployments/investigating best practices. 

The stringency of the nag pack doesn't make sense for some elements of the infrastructure as they are now (mostly related to IAM permissions stuff since we aren't exactly 100% sure about what level of permissions everything is going to need yet), but it's helpful for double-checking when I'm adding new elements to the infrastructure that nothing is horribly wrong and it also helps for learning best practices as I'm also learning as I go.

Can remove it in formal PRs if preferred

src/main.ts
```diff
@@ -36,28 +55,18 @@ const scraperTableStack = new ScraperTableStack(app, 'scraper-table-stack', {
...
   env: devEnv,
   scraperTable: scraperTableStack.scraperTable,
   uuidTable: uuidTableStack.uuidTable,
+  scrapers: scraperData.Items,
 });
 
 //makes the scraper stack provision after populator stack
 //this is to allow populator to pre-populate the uuid ddb table
-scraperStack.addDependency(populatorStack);
+// scraperStack.addDependency(populatorStack);
```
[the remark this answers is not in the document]
jiedong111: Yea, this can be removed

src/main.ts
```diff
@@ -11,18 +17,31 @@ const devEnv = {
   region: process.env.CDK_DEFAULT_REGION,
 };
 
+//to-do:
+//generate interfaces/typing for scraperData
+//reflect these changes in scraperStack
+const scraperData = JSON.parse(fs.readFileSync(path.join(__dirname, 'taskScripts/uuid-table-data.json'), 'utf-8'));
```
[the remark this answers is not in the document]
jiedong111: > Judging by how scraperData is used below, perhaps scraperMetadata is more appropriate?

Changing it to metadata makes sense.

> Are we reading a json file purely as an intermediate solution for now?

This is apparently a controversial topic, but a CDK course by Matt Bonig that I've been watching went over dealing with importing external data into your CDK stack. What we're doing seems to align with best practices with CDK, in that we are sourcing the data outside of the CDK constructor itself so we aren't breaking the contract of a constructor, and having the data within our Git history works well in validation especially once we implement CI/CD so we have an active history of whatever data we are importing into our stack and we can audit it. 

I wish I could share the video with you more easily but it's a paid course, will look more into this though and curious about your opinions on it.

src/stacks/NetworkStack.ts
```diff
@@ -0,0 +1,224 @@
...
+    // Route: '/user/logout'
+    const userServiceLogoutRoute = userServiceUserRoute.addResource('logout');
+    const logoutUserLambda = new Function(this, 'LogoutUserLambdaFunction', {
+      runtime: Runtime.PYTHON_3_12,
+      functionName: 'LogoutUser',
+      code: Code.fromAsset('src/gatewayRoutes'),
+      handler: 'logout-user.main',
+      timeout: Duration.seconds(300),
+    });
+    userServiceLogoutRoute.addMethod('GET', new LambdaIntegration(logoutUserLambda));
```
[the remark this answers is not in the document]
jiedong111: Yep, was thinking along the same lines. Wonder if it makes sense to make a Method L3 construct or something. That way we might be able to generate the function, apply auth, and add the method to the proper route through just generating those constructs as needed

src/scrapers/scraper4.py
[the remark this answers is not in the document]
jiedong111: I was imagining these `scraper[i].py` functions as mocks of actual scraper files. With this approach our orchestration layer that's running the scrapers is just EventBridge through cron rules. As it is now it's just a dummy file that spits out a mock object but in our next steps I'd like to actually run a scraper in AWS and change the validation file/scraper files in order to reflect that

src/taskScripts/populateTable.ts
jiedong111: Yep, this is once a minute in AWS cron which is apparently slightly different from how normal cron rules look

jiedong111:
> I would also like the build failure to get fixed as well. Since this is based on the `projen` branch, depending on how we want to execute the repo split/re-org we may want to merge the PR into `projen` so we can view the overall changes as well

The last time I checked the build failure was related to snapshot testing resulting in a mutation of the snapshot, I'll double-check and see if I can resolve it without it being too much of a headache, otherwise I'll see if what the build is checking makes sense for where we are right now

jiedong111:
Resolved the build check, needed to run the snapshot test once locally to have the artifact to commit and check against during the build process

src/taskScripts/populateTable.ts
nattiechan: I am not sure if we want to do every minute, even when we are just testing things? I almost wonder if there is an option to default 2x day like we decided but have the option to change the cron rules

nattiechan:
So @jiedong111 is the plan for this iteration is to have dummy scraper to unsure things work as expected before we put actual scrapers in them?

src/main.ts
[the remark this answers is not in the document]
nattiechan: No that's fine, I am just wrapping my head around what is temporary and what is not

src/taskScripts/populateTable.ts
jiedong111: I set it to 1 minute because it's useful for immediate testing of the Lambda functions when you're running `cdk watch --all --hotswap` to instantly swap Lambda defs in and out, but it is short to the point where you'd probably have to tear down `scraper-stack` if you're not working for any significant amount of time. 

That's not a problem at this point, but if we want a more stable deployment for a dev environment where things are actually persisting (DDB tables, RDS etc.) we can adjust the cron rules. After deployment, you can manually adjust the cron rules from EventBridge for each scraper from the console, but might be more efficient to just change the definition in IaC and re-deploy the scraper stack at the point. `scraper-stack` shouldn't have anything that is stateful or persisting data between deployments so tearing down and re-deploying between tests shouldn't be an issue in dev.

I'm thinking that we set up a more persistent test environment in a dev AWS account at the same time we set up CI/CD.

jiedong111:
> So @jiedong111 is the plan for this iteration is to have dummy scraper to unsure things work as expected before we put actual scrapers in them?

@nattiechan  Yea, this is basically validating that this workflow works if our scraper files are going to send and build an object with this shape:
```
    data = {
        "bank_uuid": uuid,
        "time_stamp": timeStamp,
        "hash_list": [
            {
                "hash_content": f"{timeStamp}content0",
            },
            {
                "hash_content": f"{timeStamp}content1",
            },
            {"hash_content": "initial hash"},
        ],
        "run_status": "success",
        "run_uuid": f"{uuid_util.uuid4()}-run",
    }
```
    
to the data aggregation queue. 

If it's fine that our scrapers are building an object with that structure, then we can adjust a few of our current scrapers to align with that, containerize them, and insert them into this workflow and see if things are working as expected

src/scrapers/scraper4.py
G-Weaver, line 1: Should we separate this into to two different functions? One to grab data and one to send data to SQS. While reading on this topic, it appears that separating these functions is best practice. What do you think?

src/scrapers/scraper4.py
```diff
@@ -0,0 +1,102 @@
...
+
+    # Constructs the scraper data object
+    # In this case, the hash_list which would be an md5 hash is substituted with a hardcoded list of 3 objects
+    # Hash content is either the timestamp prepended to content0 or initial hash
+    # This allows us to test the notify and the not notify paths in the same event
+    # We would expect 2 notifications to reach the user if the workflow is working properly while 1 event is disregarded
+    # We'd expect stream-reader.py to push 2 events to the queue basically
+
+    data = {
+        "bank_uuid": uuid,
```
G-Weaver, line 78: Is there a way to consistently map banks to a uuid, or are these going to be changing? (I haven't worked with uuid's at all prior to this)

src/stacks/NetworkStack.ts
```diff
@@ -0,0 +1,224 @@
...
+      'UserServiceGateway',
+      {
+        defaultCorsPreflightOptions: {
+          allowCredentials: true,
+          allowOrigins: Cors.ALL_ORIGINS, // or specify your own list of origins
+          allowMethods: Cors.ALL_METHODS, // or specify your own list of methods
+          allowHeaders: ['Content-Type', 'X-Amz-Date', 'Authorization', 'X-Api-Key'], // customize the headers as needed
+        },
+        deployOptions: {
+          stageName: 'dev',
```
G-Weaver, line 43: This should be made variable for different environments eventually. Not sure if it's worth setting up at this stage though. May be worth adding a comment/creating a ticket so we don't forget later.

src/stacks/ScraperStack.ts
```diff
@@ -21,6 +21,7 @@ import { ScraperValidator } from '../constructs/Validator';
 export interface ScraperStackProps extends StackProps {
   uuidTable: ITableV2;
   scraperTable: ITableV2;
+  scrapers: any;
```
G-Weaver, line 24: Is this a variable type, or is any just used as a placeholder here?

G-Weaver (commented):
Looks really good! I think we should create a ticket to add error handling to some of these functions in the future.

src/scrapers/scraper4.py
jiedong111: I think that would make sense as well, Lambda functions should generally be as simple and single use as possible. 

The only problem would be how we choose to move data from one function to the next. We could have one lambda function trigger another but I would generally rather avoid that since it could lead to nasty things with looping invocations if we aren't careful. We could put another queue in-between, which isn't terrible but we'd have another set of queues and DLQs to manage.

We could also package the `grab data` and `send data` step into a small step function workflow which would allow the first `grab data` Lambda function to pass context to the `send data`. Then `send data` could pass to the data aggregation queue as it's doing now. I think this is what I'd be most inclined to do if we find that it's critically important to separate the concerns of the grabbing and sending data steps

src/scrapers/scraper4.py
jiedong111: I think the uuids for each bank would be populated once when we do our initial population of the `uuid-table` then we would do as we're doing here. We scan the `uuid-table` for attributes matching the plain text `bank_name` that we're passing to each individual Scraper construct during their instantiation with the data in `data.json` which establishes which bank the individual scraper file is scraping for

src/stacks/ScraperStack.ts
jiedong111: This is a placeholder for now, I think I made a note of it in some other part of the code-base to make an interface for the `scrapers` object we're passing into `ScraperStack`. I'm not sure about the exact final data model of the object would be so I was putting it off, it's currently an array of the Items in `data.json`.

I'll make a note of it here as well

src/stacks/NetworkStack.ts
jiedong111: Makes sense, I'll make a comment

.DS_Store
nattiechan, line 1: Looks like you introduced this [here](https://github.com/sgoldfarb2/scraper-infra/commit/a658298dd443ea7af6e07aaaa45702bab07ce87b) may still need to manually remove this from the PR.

If you cannot manually delete the file, you may need to squash and reset some of your commit.

More details [here](https://github.com/sgoldfarb2/scraper-infra/pull/22#pullrequestreview-1794708292). In step 3 you are probably looking at `git reset --soft HEAD~8`

src/.DS_Store
nattiechan, line 1: same here - this needs to be removed

My understanding is that `.DS_Store` is a mac-specific thing that can get pesky sometimes

src/constructs/Notifier/notification.py
```diff
@@ -17,4 +17,5 @@ def main(event, context):
     print("Notify users!")
     print("Email?", email)
     response = sns_client.publish(TopicArn=topic_arn, Message=f"{message}-{timeStamp}")
+
```
nattiechan, line 20: When there is a red "stop" sign at the end below, generally it means it is not having a new line at the end even though you seem to have one. Check this line and see if there is any whitespace in there, and same for other Python files that have the same thing

src/main.ts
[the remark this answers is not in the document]
nattiechan: I think once we finalize how this should look like let's remove this.

src/main.ts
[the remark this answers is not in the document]
nattiechan:
```suggestion
const scraperMetadata = JSON.parse(fs.readFileSync(path.join(__dirname, 'taskScripts/uuid-table-data.json'), 'utf-8'));
```

src/scrapers/scraper4.py
nattiechan: I would advocate making lambda functions as single use as possible as well

src/taskScripts/populateTable.ts
nattiechan: Okay. We should definitely make a note that when we transition to actual scrapers this CANNOT be at 1 min interval.

src/constructs/Notifier/notification.py
jiedong111: Should be resolved in latest commits

src/taskScripts/populateTable.ts
jiedong111: Note added in `populateTable.ts`

src/scrapers/scraper4.py
jiedong111: That sounds good, I'll look into transitioning to running express step functions for the first few Lambda functions. It's possible we can place the entire scraper mechanism into a step function for a higher level of orchestration. Do we want this resolved prior to merging this PR or would it make more sense for a future one?

Image attached mocking how this would change our data flow:

current to stepfunctions

nattiechan: I do agree the one to the right is what we want. Ultimately the scraper themselves should be isolated machines, and whatever functions that come before/after should be consider part of the scraper operation. This very much resembles how we had things in Kubernetes and I think it should work okay

PR #26, opened 2023-12-21
Feature/gh 119 add bank of aruba scraper

kevin-06-huang:
## Describe your changes

- add Central Bank van Aruba scraper
- Resolves #119

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# Steps to test code locally (provide context as needed)

- Navigate to where the repository is locally
- Run `python ./scrapers/<your-scraper-file>`

# Checklist:

 - [x] Changes have been properly documented via comments and/or readme
 - [x] Changes generate no new warnings / errors
 - [x] Changes passes all new and already existing tests
 - [x] Any dependency changes have been highlighted in PR description


scrapers/bank_of_aruba.py
```diff
@@ -0,0 +1,40 @@
...
+        page = browser.new_page()
+        page.goto(CIRCULAR_URL)
+
+        session = requests.Session()
+
+        page.locator("#article-1").click()
+
+        pdf_link = page.locator("#articleBody a").all()[0].get_attribute("href")
+
+        if not pdf_link:
```
jiedong111, line 21: Here you're asserting that the only error that would cause the scraper to stop functioning is an invalid PDF link. Take a look in `utils.py` and see if there's something you could be using here that's more robust

jiedong111 (changes requested):

scrapers/bank_of_aruba.py
kevin-06-huang: @jiedong111 you mean `def validate_and_get_download_link(base_url, download_link, check_content_pdf=True)` right?

PR #28, opened 2023-12-21
feat: add central bank of west african states scraper

mlamchamkee:
Resolves: #73
https://github.com/sgoldfarb2/scraper/issues/73


scrapers/central_bank_of_west_african_states.py
```diff
@@ -0,0 +1,77 @@
...
+    """
+    with sync_playwright() as pw:
+        browser = get_browser(pw)
+        page = browser.new_page()
+        page.goto(f"{base_url}{circular_path}")
+
+        # Since the PDFs are linked in each page individually,
+        # need to first find the page links
+        page_links = page.locator("div.elemReg > a").all()[0:count]
+        links = [page_link.get_attribute("href").rstrip() for page_link in page_links]
```
nattiechan, line 38: So looking at the page itself one thing stood out to me - for some reason they publish like 5 million things on one day. If you scroll through the page you will notice that there are new publications on the latest date on page 2 as well. This is a bit of a red flag for me because that means we need to add something to the scraper to ensure we are getting the number that we want (and perhaps in the future, we may need to have this scraper to not be dependent on the count)

Can we add some logic here to:

- have the capability to go to other pages
- Compare say, 2-3x of the circulars that we need by date, and ensure we are capturing the n latest ones

nattiechan (changes requested):

scrapers/central_bank_of_west_african_states.py
mlamchamkee: This is a great callout!  I actually noticed that and took the view that when the user requests n circulars, that it meant n PDFs.  Are we taking a different view that n would correlate to dates and in this case if the user request the latest 2 circulars, the expected behavior would be to send 9 PDFs for Oct 20 and 1 for Oct 10?

The date is in a `span`, which is a child of the `a` tag.  So, the above should be manageable, should we decide to go that route.

nattiechan: > I actually noticed that and took the view that when the user requests n circulars, that it meant n PDFs

That is assuming your bank does not separate the actual circular memo from things like guidance, details that are meant to be an "appendix" or "attachment" to the actual notice. See https://github.com/sgoldfarb2/scraper/issues/100 for what I mean. I _think_ you are in the clear with this bank but please double check.

> Are we taking a different view that n would correlate to dates and in this case if the user request the latest 2 circulars, the expected behavior would be to send 9 PDFs for Oct 20 and 1 for Oct 10?

Yes, assuming all 10 PDFs are for different topics.

